PR #131, opened 2012-11-25
Fix compatiblity with PostgreSQL.

LEW21:


DavidIQ:
This pr is pretty old and against a branch that is no longer in use.  As such...closing.

PR #141, opened 2015-02-03
Small change to allow mpv server to validate unapproved MODs

DavidIQ:


controller/download.php
```diff
@@ -267,7 +267,7 @@ protected function is_auto_validator()
 			$dns_ipv4 = dns_get_record($data['host'], DNS_A);
 			$dns_ipv6 = dns_get_record($data['host'], DNS_AAAA);
 
-			if ($this->dns_matches_user_ip($dns_ipv4) || $this->dns_matches_user_ip($dns_ipv6))
+			if ($this->dns_matches_user_ip($dns_ipv4) || $this->dns_matches_user_ip($dns_ipv6) || (isset($data['lb']) && $dns_ipv4 == $data['lb']))
```
michaelcullum: I'd rather configure the actual (www2) server ip here (and check if that's the ip), rather than just checking if `host` resolved to the `lb` ip. By doing it the way it's done here, it will only check if mpv.phpbb.com resolves to an lb, and if that's true, return true, so there is no checking on the request's ip at all (current implementation is a permissions bypass sec vuln). The lb ip will be changing in the future.

michaelcullum:
:+1:

PR #146, opened 2015-06-01
[ticket/661] Provide a way to get list of approved contributions

lavigor:
Users really need the official feature to get the list of validated contributions.

The feature was created for further integration with other tools such as Upload Extensions.

Implemented features:
* Ability to get a list of contributions of a special type.
* Ability to get a full list of all contributions grouped by their types.
* Ability to catch different contributions that have the same name: contributions are caught by their IDs, not by their names.

Further possible improvements to do:
* Group contributions by minor phpBB version numbers (3.1, 3.2 and so on), not by full version numbers (3.1.2, 3.1.3 etc.).
* Allow list downloads for specific phpBB versions (only for 3.1, only for 3.2 etc.).
* Add short text descriptions for contributions.
* Add checksums.
* Allow limited downloads (by small portions).

[CUSTDB-661]

### DIFF
--- a/config/routes/general.yml
+++ b/config/routes/general.yml
@@ -18,6 +18,12 @@ phpbb.titania.all_contribs:
     pattern: /contributions/all
     defaults: { _controller: phpbb.titania.controller.all_contribs:display_all_contributions }
 
+phpbb.titania.get_all_contribs:
+    pattern: /contributions/get/{type}
+    defaults: { _controller: phpbb.titania.controller.all_contribs:get_all_contributions, type: all }
+    requirements:
+        type: all|mod|extension|style|converter|official_tool|bbcode|bridge|translation
+
 phpbb.titania.author:
     pattern: /author/{author}/{page}
     defaults: { _controller: phpbb.titania.controller.author:base, page: details }
--- a/controller/all_contribs.php
+++ b/controller/all_contribs.php
@@ -90,4 +90,18 @@ class all_contribs
 		$data['sort']->set_url($this->helper->route('phpbb.titania.all_contribs'));
 		$this->template->assign_var('U_CANONICAL', $data['sort']->build_canonical());
 	}
+
+	/**
+	 * Get the list of contributions in JSON format.
+	 *
+	 * @param string $type	Contrib type URL identifier.
+	 *
+	 * @return \Symfony\Component\HttpFoundation\JsonResponse
+	 */
+	public function get_all_contributions($type)
+	{
+		$contribs = \contribs_overlord::get_contribs_list($type);
+
+		return new \Symfony\Component\HttpFoundation\JsonResponse($contribs);
+	}
 }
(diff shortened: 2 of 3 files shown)


lavigor:
> We review all pull requests sent here.

@DavidIQ, could you please let me know whether you've seen this PR?
I really think that board administrators need this feature and I am also aware of the fact that I'll need to improve the code (and add some new features from the list above).

DavidIQ:
Checksum should already be in the attachments table and probably a good thing to use to ensure proper download integrity before attempting to apply a contribution to a board.  A bad download can render a forum unusable.  I've looked through the code and, assuming you tested this on an actual instance of Titania, it looks ok but I'd also like @prototech to look it over as he either had some WIP or additional comments on this planned feature.

michaelcullum:
This (A REST API for titania) has already been built, it's just in the format of a statis repository for composer integration in 3.2 (still serves a json response, just in a particular format).

For the moment we don't want to add another REST API to the CDB as it will be deprecated in a few months anyway, but by that point people may rely on it so we will be in a position in which we have to maintain two REST APIs.

It may also encourage integrating the cdb (tools you mention such as upload extensions) into phpBB installs via extensions in 3.1 which is not something we want people to be doing (There are a large number of non-obvious security concerns in addition to the fact that it's extremely unlikely, if not impossible, to be done in the clean fashion that we are working on for 3.2).

lavigor:
@michaelcullum, as I understood from your reply, do you mean that there should be NO people who need tools integrating with CDB like Upload Extensions because of security reasons?

Do you also mean that there will be no integration until the end of the year when phpBB 3.2 is planned to be released? Where can I find any news about the mentioned REST API?

If you say that you do not like extensions that use information from the CDB I can collaborate in the development of the official tool. I've spent enough time on simplifying the process of installing extensions that can be secure (see Upload Extensions 3.1.2).

michaelcullum:
I'd certainly be interested in seeing how you've done it. Do you use IRC at all?

lavigor:
> I'd certainly be interested in seeing how you've done it.

If you are talking about Upload Extensions, I've done it myself, most of the code is handmade (other part is some libraries).
If you're talking about this PR, I've tested it on a copy of Titania on my board.

> Do you use IRC at all?

No, but I'm registered in some social webs. The links can be found on my personal website.
You can [PM me](https://www.phpbb.com/community/ucp.php?i=pm&mode=compose&u=1406196) on phpbb.com for contact information.

prototech:
The mentioned API was never deployed due to there being no actual working code in the core to make use of it at the time. As such, it's currently not in a working state. I'll get that working pretty soon for general usage and to make it available for testing the core implementation of the installer.

DavidIQ:
I'm reopening this because there is no documentation or trace of the API mentioned by @michaelcullum 

Can you provide more details? If not then this needs revisiting unless @prototech has something.

michaelcullum:
@DavidIQ Cesar never converted it over to 3.1 titania and he was going to do it but he's been away. When he get's back he'll re-add it to 3.1 titania so that the core integration work can continue.

PR #149, opened 2015-09-10
Composer repository

Nicofuma:
Fix the old implementation in order to make titania acting as a valid composer repository.

A few important things:
- Any package in the 'composer compatible types' list and with a composer.json file is added to the index
- Only validated packages should be present
- We use the original composer.json file instead of building a new one. This way we preserve the original name, dependencies etc...
- The `phpbb/phpbb` requirement is automatically added (if it's not already present) and copied from `['extra']['soft-require']['phpbb/phpbb']`

Note1: It is waiting for a PR against the core to enable the ResonseListener. Otherwise the endpoints to get the json files doesn't work (didn't sent it yet because JIRA is broken)


Nicofuma:
@VSEphpbb the issues should be solved now

VSEphpbb:
Yep. Looks like package list is rebuilt on contrib approval :+1: 

It is doing some weird stuff with the Version check JSON data, ie: host: localhost

Nicofuma:
that's intended, I think

includes/objects/revision.php
```diff
@@ -706,19 +707,21 @@ public function update_composer_package($mode = 'add')
...
 
 		if ($mode == 'add')
 		{
-			$package_manager->add_release($this->revision_version, $this->attachment_id, true);
+			$package_manager->add_release(json_decode($this->revision_composer_json, true), $this->attachment_id, true);
 		}
 		else
 		{
-			$package_manager->remove_release($this->revision_version);
+			$package_manager->remove_release($this->revision_composer_json['version']);
```
prototech: This is still a string and not an array. Last error here https://tracker.phpbb.com/browse/CUSTDB-697 refers to this.

includes/tools/composer_package_manager.php
```diff
@@ -325,25 +334,21 @@ public function get_packages_group($write_map)
...
-			'version'	=> $version,
-			'dist'		=> 
-				array (
-					'url'	=> titania_url::build_clean_url('download', array('id' => $attachment_id)),
-					'type'	=> 'zip',
-			),
+		$controller_helper = phpbb::$container->get('phpbb.titania.controller.helper');
+
+		$composer_json['dist'] = array(
+			'url'	=> $controller_helper->route('phpbb.titania.download', array('id' => $attachment_id)),
```
prototech: There's a possibility of the user's sid being included in here. Use the path helper to strip that parameter.

controller/composer.php
```diff
@@ -0,0 +1,63 @@
...
+	* @return \Symfony\Component\HttpFoundation\Response
+	*/
+	public function serve_file($filename)
+	{
+		if (strpos($filename, '..') !== false)
+		{
+			throw new http_exception(404, 'NO_PAGE_FOUND');
+		}
+
+		$filename = \titania::$root_path . 'composer/' . $filename . '.json';
```
prototech: Use the `%phpbb.titania.root_path%` parameter instead of `\titania::$root_path`.

prototech:
The version check data is based on the generated URL for the local installation. It's not hard-coded to point to phpBB.com.

includes/objects/revision.php
Nicofuma: fixed

includes/tools/composer_package_manager.php
Nicofuma: fixed

controller/composer.php
Nicofuma: fixed

prototech:
@michaelcullum Please update the community board now that the required core PR has been merged.

michaelcullum:
First of all it needs to be decided for sure that we are reversing the very concious decision in 3.1 to have a soft requirement differ from the hard composer requirement.

Nicofuma:
@phpbb/development-team @VSEphpbb @Derky @paul999 @DavidIQ ^

DavidIQ:
@michaelcullum can you elaborate on what you just said?

nickvergessen:
We only moved to soft-require, because we didn't have phpbb set up on packagist back then?

Derky:
Yeah that's what I thought as well. (and soft-require isn't part of the composer spec, it has just been made up for the reason Joas mentioned IIRC)

DavidIQ:
Then there isn't really anything to hold this back, unless @michaelcullum has some additional information that would stop this from going any further.

paul999:
@michaelcullum We are kinda waiting on this, can you please comment or can we otherwise merge it?

Nicofuma:
By the way, @paul999, if someone want to use the packages.json file he needs to support the whole composer format because we may change it in the future (it could be needed if there if too much packages to serve). Also, this file may serve non phpBB package (a set of approved packagist packages per example).

In composer source code, the implementation related to the format we use is this one: https://github.com/composer/composer/blob/master/src/Composer/Repository/ComposerRepository.php (we don't use any provider for now, but it may change).

https://getcomposer.org/doc/05-repositories.md#composer

paul999:
@Nicofuma I guess we can put that in the Titania documentation? :P

Nicofuma:
That's a classic implementation following these "specs" https://getcomposer.org/doc/05-repositories.md#composer ;)

paul999:
Ok. However we should still document somewhere in the titania documentation that we follow the standard specs?

Nicofuma:
There is a documentation? =D I guess yes...

michaelcullum:
Titania has no documentation really?

Sorry I haven't gotten around to responding to this yet, I'll respond to this PR tomorrow (I'm on my phone right now). This PR is not just waiting on me to comment on it as it also requires me to do some stuff to our phpbb installation too before it can be merged.

paul999:
@michaelcullum this has some sort of priority as stuff (/users) is waiting on it.

michaelcullum:
Sure, it's top of my phpBB to do list, but I've been busy with real life this last week, the exceptions being when websites keep dying and I've had to sort out stuff with OSL.

Elsensee:
I'm not that much into composer but I got a question:
What happens if someone wants to install an extension which has phpbb as a requirement. Will the extension still be put in the correct place (like it's done when you set the extension as a requirement in your phpbb installation)?

Nicofuma:
yep

paul999:
@michaelcullum Bump

DavidIQ:
This will be merged by 11 AM UCT-4 today unless there are any actual objections that don't include "I'll reply by x time/day".

DavidIQ:
Command/feature to import already approved extensions to come later.

michaelcullum:
Sure, because <3 hours is plenty of time. Merging this PR will also break phpBB if you actually read it, it requires some changes to phpBB, which also need me to make them. ;)

DavidIQ:
@michaelcullum  < 3 hours???? You've had over _20 days_ to say what you needed to say and have not said anything useful whatsoever.

PR #151, opened 2015-10-07
Do not allow submission of new MODs

VSEphpbb:
This should make it so people can no longer submit new MODs (the Modification option will no longer be visible to them when creating a new contribution, but they can still submit new revisions to validated MODs). @prototech

### DIFF
--- a/includes/types/mod.php
+++ b/includes/types/mod.php
@@ -102,7 +102,7 @@ class titania_type_mod extends titania_type_base
 		{
 			// Can submit a mod
 			case 'submit' :
-				return false;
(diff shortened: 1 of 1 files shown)


DavidIQ:
This change will also prevent MOD updates, which are still allowed until next month.  See https://www.phpbb.com/about/release/

VSEphpbb:
@DavidIQ Are you sure? I was able to submit a new revision in testing.

DavidIQ:
Easy to revert anyways.  Let's just merge it.

PR #153, opened 2015-10-09
Composer repository

michaelcullum:
New composer repository PR
customisation-db#149

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -59,9 +59,6 @@ class config extends \phpbb\titania\entity\base
 			'language_path'				=> array('default' => $this->ext_root_path . 'language/'),
 			'table_prefix'				=> array('default' => 'cdb_'),
 
-			// Unique Composer vendor name.
-			'composer_vendor_name'		=> array('default' => ''),
-
 			// Increment the user's post count?  Array of the post_types for which we will increment the post count
 			'increment_postcount'		=> array('default' => array(TITANIA_SUPPORT)),
 
--- a/config/controllers.yml
+++ b/config/controllers.yml
@@ -352,3 +352,10 @@ services:
             - @phpbb.titania.display
             - @phpbb.titania.tracking
             - @phpbb.titania.sort
+
+    phpbb.titania.controller.composer:
+        class: phpbb\titania\controller\composer
+        arguments:
+            - @user
+            - @phpbb.titania.controller.helper
+            - %phpbb.titania.root_path%
--- a/config/routes/general.yml
+++ b/config/routes/general.yml
@@ -60,6 +60,10 @@ phpbb.titania.contribution.id:
     requirements:
         id: \d+
 
+phpbb.titania.composer:
+    path: /composer/{filename}.json
+    defaults: { _controller: phpbb.titania.controller.composer:serve_file }
+
 phpbb.titania.category:
     pattern: /{category1}/{category2}/{category3}/{category4}
     defaults: { _controller: phpbb.titania.controller.index:display_category, category2: '', category3: '', category4: '' }
@@ -69,3 +73,4 @@ phpbb.titania.legacy:
     defaults: { _controller: phpbb.titania.controller.legacy_rerouter:redirect }
     requirements:
         path: .*
+
--- a/controller/composer.php
+++ b/controller/composer.php
@@ -1,0 +1,67 @@
+<?php
+/**
+*
+* This file is part of the phpBB Customisation Database package.
+*
+* @copyright (c) phpBB Limited <https://www.phpbb.com>
+* @license GNU General Public License, version 2 (GPL-2.0)
+*
+* For full copyright and license information, please see
+* the docs/CREDITS.txt file.
+*
+*/
+
+namespace phpbb\titania\controller;
+
+use phpbb\exception\http_exception;
+use Symfony\Component\HttpFoundation\BinaryFileResponse;
+
+class composer
+{
+	/** @var \phpbb\user */
+	protected $user;
+
+	/** @var \phpbb\titania\controller\helper */
+	protected $helper;
+
+	/** @var string */
+	protected $titania_root_path;
+
+	/**
+	 * Constructor
+	 *
+	 * @param \phpbb\user $user
+	 * @param helper $helper
+	 */
+	public function __construct(\phpbb\user $user, \phpbb\titania\controller\helper $helper, $titania_root_path)
+	{
+		$this->user = $user;
+		$this->helper = $helper;
+		$this->titania_root_path = $titania_root_path;
+	}
+
+	/**
+	* Serve composer's files.
+	*
+	* @param string $filename	Filename to serve (without extension)
+	* @return \Symfony\Component\HttpFoundation\Response
+	*/
+	public function serve_file($filename)
+	{
+		if (strpos($filename, '..') !== false)
+		{
+			throw new http_exception(404, 'NO_PAGE_FOUND');
+		}
+
+		$filename = $this->titania_root_path . 'composer/' . $filename . '.json';
+
+		try
+		{
+			return new BinaryFileResponse($filename, 200);
+		}
+		catch (\Exception $e)
+		{
+			throw new http_exception(404, 'NO_PAGE_FOUND');
+		}
+	}
+}
--- a/controller/helper.php
+++ b/controller/helper.php
@@ -44,9 +44,9 @@ class helper extends \phpbb\controller\helper
 	/**
 	* {@inheritDoc}
 	*/
-	public function render($template_file, $page_title = '', $status_code = 200, $display_online_list = false)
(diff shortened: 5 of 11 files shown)


DavidIQ:
Could we not sit on this again?  The last pr sat without action for some 29 days before it was merged and then quickly reverted.  Since apparently you are the only one that knows what needs to be changed in phpBB (not even our own developers seem to know) then this is still waiting for you to do whatever changes are needed on the phpBB side.

michaelcullum:
'Not even our own developers' - Well, it was a developer who told me it needed doing? It's also detailed in the previous PR....

DavidIQ:
@Nicofuma (one of our developers) had mentioned he would have made the change already if he knew what it was, which I'm having a hard time finding what the change is in the previous PR.  Here are the 4 messages you posted to the PR:
>First of all it needs to be decided for sure that we are reversing the very concious decision in 3.1 to have a soft requirement differ from the hard composer requirement.
_______________________________________________
>Titania has no documentation really?

>Sorry I haven't gotten around to responding to this yet, I'll respond to this PR tomorrow (I'm on my phone right now). This PR is not just waiting on me to comment on it as it also requires me to do some stuff to our phpbb installation too before it can be merged.
_______________________________________________
>Sure, it's top of my phpBB to do list, but I've been busy with real life this last week, the exceptions being when websites keep dying and I've had to sort out stuff with OSL.
_______________________________________________
>Sure, because <3 hours is plenty of time.... Merging this PR will also break phpBB if you actually read it, it requires some changes to phpBB, which also need me to make them. ;)

The only other thing I could find in that PR that might be of relevance is this line from @Nicofuma:
>Note1: It is waiting for a PR against the core to enable the ResonseListener. Otherwise the endpoints to get the json files doesn't work (didn't sent it yet because JIRA is broken)
Not sure if that's what you're referring to that was detailed in PR #149.

michaelcullum:
As I said, a developer posted it in the old PR... Cesar...

DavidIQ:
Must be this one-liner then:
>Please update the community board now that the required core PR has been merged.

When do you plan on updating the board then?

paul999:
@DavidIQ @michaelcullum bump

paul999:
@michaelcullum Can we at least get something going here? Promising that you will look at it next day and not looking at it for weeks doesn't help really...
/cc @DavidIQ

michaelcullum:
@paul999 I've spoken to David about it.

paul999:
Well, lets now do something more as just speaking to each other and get this online?

michaelcullum:
As in, he knows why it's not done and there's nothing that can be done as it's waiting on things outside our control.

michaelcullum:
Also, this now has conflicts. @Nicofuma

lavigor:
Is there any ETA when this feature will be available on phpbb.com?
And could you tell me the direct link for the list of extensions that I can add to Upload Extensions?
Because this feature is the only reason why this extension was not released for more than 100 days.

michaelcullum:
@lavigor Now, see https://github.com/phpbb/customisation-db/pull/153/files#diff-88a0179a60f880a312021d194ae3123aR63 and it's a composer repository.

prototech:
It'll be fully available by tomorrow. The relevant file for you will be https://phpbb.com/customise/db/composer/packages-extension.json which will have list of includes https://getcomposer.org/doc/05-repositories.md#includes These includes will contain the list of packages https://getcomposer.org/doc/05-repositories.md#packages

Nicofuma:
But you should support the full composer spec on that matter (including providers) because we may have to change this behaviour in the future (but it will still be a valid composer repository)

lavigor:
OK, thank you for your replies!

> But you should support the full composer spec on that matter (including providers)

@Nicofuma, I think that I can support the repository only 'as is' because I haven't found any PHP library handling composer stuff without command line interface.
So I'll need to write a wrapper for the current structure of the file and update the extension when such structure is changed.

@prototech, thank you very much for the feature updates. As I can see, full structure of composer.json files will be available in the composer repository (including full names, descriptions etc.), right?

Nicofuma:
You can just look (and maybe reuse) composer source code.

prototech:
@lavigor This is up now :) https://www.phpbb.com/customise/db/composer/packages-extension.json

lavigor:
@prototech Great! Really great!
Thank you!:+1:

lavigor:
@prototech Could you please also add md5 hashes to the download links?
Or is it against Composer structure?

prototech:
Composer allows a sha1 checksum to be provided in the dist array as shown below, so we'll be switching Titania to using sha1 checksums instead of md5.
```
    "dist": {
        "url": "https://www.phpbb.com/customise/db/download/111761",
        "type": "zip",
        "shasum": "f6949a8c7d5b90b4a698660bbfb9431503fbb995"
    }
```
Providing the contrib URL should also not be a problem. We're currently discussing where to place this. Most likely it'll be in the homepage property.

lavigor:
@prototech OK, very glad to hear that.
However, I'm experiencing difficulties with the implementation for the list of extensions with the current approach to repository creation.

I think that all packages should be uploaded to every board's cache, otherwise it is not possible to perform search/sorting/other related actions, am I right?

@Nicofuma what do you think about decreasing performance when all files are uploaded (and all of them are loaded to the RAM) if the repository is using providers?

I thought that it would be better to have an API (such as the one that Wordpress has) like that:
packages/providers => list of common variables (name, version and description, possibly download link too) => separate files for each package with other details.

Or maybe I haven't understood the way of handling Composer repositories and there's a way to optimize search performance without the need to download and cache all CDB information?

Nicofuma:
It's not an issue for now because their isn't a lot of data.

About the API, it could be an idea, but that's another work and it has to be designed with a lot of things in mind (including user land caching because you don't want to make an http on each catalog page, it's too slow especially if you have any kind of connectivity issue.)

marc1706:
I think splitting the data up is a rather good idea. There should IMHO be a main file with just the common variables. If one does click on details regarding the package, another file with the full extension data will be fetched from our servers.

Nicofuma:
But that's not possible, not with composer.

There is too things here:
- Composer
- The others usages

The first (and main goal) of this PR was not to provide an extension list but a composer repository and we had 2 solutions:
- The one currently implemented
- The providers (to sum-up, we publish a list of package name and the URL pattern to retrieve the composer.json of a given package. That's what have been implemented for packagist because they can't serve a 1GB file)
The second solution could look better but not necessarily in our case:
- With the first solution we are serving 5 files (could be reduced) for a total of less than 250KB
- The second solution involves, at least, one HTTP call to retrieve the list of packages and then one HTTP call per required package. It can grow up quickly.
- With the second solution you have a better control on the cache (when you validate an extension it will not invalidate all or almost all existing file) but I don't we don't have to worry about that here for 2 reasons:
  - There isn't that much activity around extensions validation
  - An extension will always invalidate 3 files, no more: packages.json, packages-extension.json and packages-extension-<number>.json (number being between 1 and 3 for now). (By the way, I think e should remove packages-extension.json and move its content to packages.json if it is possible)

To sump-up, the second solution scale better when you have thousands of packages (one version === one package) like packagist and you need more than 100MB to store them all but the first solution is better at lower scale because it involves less HTTP calls for a memory consumption which remains low.

About the extension list, we can use the list published by composer but the best is to provide a dedicated API with filter (author, vendor, package type etc...) and why not search.

lavigor:
> There should IMHO be a main file with just the common variables.

+1 I definitely agree with everything that @marc1706 has mentioned.
In fact I thought that an API can be implemented exactly in the described way.
I even tried to start that implementation in [my PR](https://github.com/phpbb/customisation-db/pull/146) several months ago.

> The first (and main goal) of this PR was not to provide an extension list but a composer repository

@Nicofuma Could you please explain the total benefits for using a Composer repository for phpBB at all?
I agree that Composer is used by Packagist (and it has benefits there), but that resource is mainly for developers.
I thought about implementing a list for *end users* that are still forced to use FTP and create proper directories to install every single extension. I do not think that most *end users* will open command line to set up Composer in order to install a phpBB extension.

As for developers, extensions are greatly located on the mentioned Packagist.
So I cannot understand the reason of having a very special phpBB Composer repository.

> the second solution scale better

The second solution is the same and wastes the same amount of resources unless useful information is added to the list of packages. Because the search and other actions cannot be performed without downloading and caching everything with this approach.

Moreover, name field is not as important as display-name field in context of phpBB (again, for *end users*).
For Packagist name field is important for search as it is the 'main name field', so it is OK there, but not in phpBB.

So I still do not know why *end users* should upload all packages of a special Composer repository instead of requesting necessary information in an API call and caching only relevant data.

>About the extension list, we can use the list published by composer but the best is to provide a dedicated API with filter (author, vendor, package type etc...) and why not search.

Is it possible to create an API just now? I really want to help people waiting for it.

lavigor:
@Nicofuma @prototech Any plans for API?
@prototech What about `shasum` and contrib URL?

PR #155, opened 2015-10-20
Fix some errors with sphinx search

Zoddo:


search/driver/fulltext_sphinx.php
```diff
@@ -182,7 +182,6 @@ public function search()
 			foreach ($result['matches'] as $data)
 			{
 				$attrs = $data['attrs'];
-				$attrs['title'] = $attrs['title_attr'];
```
Zoddo: `title` is already returned by the SQL query

PR #159, opened 2015-12-01
Allow FAQ postbody to span 100%

VSEphpbb:
The FAQ's are scrunched down to 76% to make room for profiles, but there are no profiles to display. So this is just a remnant of relying on prosilver styling/classes. There's no reason the FAQs can't span the whole width of the post content area, so this little fix allows for that.


prototech:
Use the `full` class alongside .postbody instead of the inline style.

VSEphpbb:
Thanks. Updated.

PR #160, opened 2015-12-03
Add "Validated for: version #" to download button info

VSEphpbb:
This adds to the download button info box that appears a little text:
Validated for:
phpBB 3.1.X

There are always people who think the version we display in the download button means that the contribution is ONLY for that version of phpBB. I am hoping this little addition of info makes it clearer that the version of phpBB displayed is not a rule, it's just our way of showing what version of phpBB we used when validating the contrib.

screen shot 2015-12-03 at 9 35 01 am


DavidIQ:
I highly doubt this will curb what you're intending it to curb, but it doesn't hurt anything either I guess.

PR #162, opened 2015-12-16
Improve the phpbb/phpbb requirement generation

Nicofuma:


DavidIQ:
Is there any background to this?

Nicofuma:
There is a lot of extension with a soft-require like this one: `<=3.2.*@dev` which isn't a valid constraint because `<=` and `3.2.*` both define a range. 

What I am doing here is removing the `.*` for as many case as possible (`<3.2.*`, `<=3.2.*`, `>=3.* `, ...)

VSEphpbb:
regex pattern looks good.

PR #163, opened 2015-12-16
Build a composer version of the extension package

Nicofuma:


contribution/extension/type.php
```diff
@@ -210,7 +211,7 @@ protected function repack(package $package, \titania_contribution $contrib, \tit
 			throw new \Exception('INVALID_EXT_NAME');
 		}
 
-		$ext_name = $data['name'];
+		$ext_name = $data['name'];$ext_full_name = $data['name'];
```
VSEphpbb: new line?

lavigor:
@prototech @Nicofuma Could you please explain the difference between 'manual' and 'composer' types of packages?
Will composer packages contain less files so that Composer is able to install dependencies itself, or is it just the matter of restructuring ZIP archives?
I wonder what type I should use in Upload Extensions to have no worries about possible future incompatibility.

Nicofuma:
It's just the packaging because composer expects an archive with a single root directory ('dir1/ext.php' versus 'vendor/name/ext.php')

lavigor:
OK. Thank you for the explanation.

PR #167, opened 2016-02-16
Fix hidden sub categories

VSEphpbb:
Children categories are not hidden when their parent is hidden.


VSEphpbb:
@phpbb/extensions-team

PR #181, opened 2016-07-31
Add Contributions link everywhere

Elsensee:
As per [CUSTDB-669](https://tracker.phpbb.com/browse/CUSTDB-669).


prototech:
U_MY_CONTRIBUTIONS doesn't exist outside of Titania so this change is moot.

michaelcullum:
Hm. Could we move it into a listener @Elsensee?

Elsensee:
I knew, this was too easy.

PR #183, opened 2016-07-31
Really show My contributions-link

Elsensee:
This extends the previous PR #181 to actually make it work.

### DIFF
--- a/event/main_listener.php
+++ b/event/main_listener.php
@@ -156,6 +156,18 @@ class main_listener implements EventSubscriberInterface
 
 	public function overwrite_template_vars($event)
 	{
+		if ($this->user->data['is_registered'] && !$this->user->data['is_bot'])
+		{
+			$u_my_contribs = $this->controller_helper->route('phpbb.titania.author', array(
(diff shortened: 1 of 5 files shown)


event/main_listener.php
```diff
@@ -156,6 +156,18 @@ public function startup($event)
 
 	public function overwrite_template_vars($event)
 	{
+		if ($this->user->data['is_registered'] && !$this->user->data['is_bot'])
+		{
+			$u_my_contribs = $this->controller_helper->route('phpbb.titania.author', array(
```
michaelcullum: If not in titania will this send to /customise/db/? I think not. @prototech

prototech: It won't. 😞

Elsensee: How often will you stab me if I use some kind of `str_replace($titania_config->phpbb_script_path, $titania_config->titania_script_path, ...);` here?

michaelcullum:
Is this tested?

Elsensee:
Yes.

PR #191, opened 2016-11-18
Format & fix EPV output

VSEphpbb:
EPV output is unformatted and kind of a mess:

screen shot 2016-11-18 at 12 46 27 am

This PR tries to correct it:
screen shot 2016-11-18 at 8 13 07 am


styles/prosilver/template/contributions/contribution_revision.html
```diff
@@ -129,7 +130,9 @@ <h3 class="section-name">{PAGE_TITLE}</h3>
 			<!-- IF PV_RESULTS -->
 				<p class="error">{L_PV_RESULTS}</p>
 				<hr class="dashed" />
-				{PV_RESULTS}
+				<div style="font-size:12px;">
```
DavidIQ: Shouldn't this be in a stylesheet?

VSEphpbb:
Styling comes from pro silver. So that's an inline override.

PR #194, opened 2017-01-02
Fix styling issues

VSEphpbb:
New BBCode system change's line breaks around lists. We now rely on CSS to add margin to bottom of lists.

### DIFF
--- a/assets/javascript/ajax.js
+++ b/assets/javascript/ajax.js
@@ -53,7 +53,7 @@
 		var $form = $(this),
 			$postbody = $form.parents('.postbody');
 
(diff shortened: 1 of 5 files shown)


VSEphpbb:
Just added another fix.

line breaks are looking extra big, there should only be 1 line between this link and paragraph:
screen shot 2017-01-03 at 8 25 31 am

It appears to be due to an old line of CSS that was originally added to deal with some JS plugin we don't even have anymore called "tabby" https://github.com/phpbb/customisation-db/commit/c4acb1b305bc642a92c88cbe26351e23e4ed812f#diff-551a4d811fe4fd7cb787cd4175fee436

PR #195, opened 2017-01-05
Fix output of EPV

VSEphpbb:
Fix an issue in phpBB 3.2 where EPV output, which contains CLI formatting such as `<success>` was now being encoded to HTML entities, and thus rendering `<success>`, e.g.

unspecified

### DIFF
--- a/contribution/extension/prevalidator.php
+++ b/contribution/extension/prevalidator.php
@@ -24,6 +24,7 @@ class prevalidator
 	 *
 	 * @param string $directory		Directory where extracted revision is located
 	 * @return string
+	 * @throws \InvalidArgumentException
 	 */
 	public function run_epv($directory)
 	{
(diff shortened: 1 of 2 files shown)


VSEphpbb:
Well this does not fix EPV output when EPV is run from the queue. So another solution is needed.

Why does the EPV generate all this formatting stuff here? 
https://github.com/phpbb/customisation-db/blob/3.2.x/contribution/extension/prevalidator.php

VSEphpbb:
Maybe this???

prototech:
Just go ahead and nuke those tags in run_epv()? Pretty sure they don't serve a purpose here.

VSEphpbb:
@prototech They probably don't but the problem is some tags come in on the individual messages anyways directly from the EPV, so even if we remove the hardcoded tags here, we still get some from EPV itself. So we will still need to replace them.

VSEphpbb:
Now it outputs a cleaner result. Here's an image capture from what is now posted in the queue:
screen shot 2017-01-06 at 2 34 36 pm

michaelcullum:
Ping me when you want this deployed

paul999:
I have no idea how the EPV implementation works in titania, thats something @prototech made I think. EPV has nothing changed, so it is a 3.2/titania issue I guess.

VSEphpbb:
@michaelcullum  deploy it

PR #196, opened 2017-01-07
Add 3.2.0 package

marc1706:


michaelcullum:
More changes than this needed included config files

marc1706:
Config is already up to date. Only need to merge and update phpbb3-website

DavidIQ:
The .com specific config needs updating but not for this. I'll take care of that separately.

PR #201, opened 2017-01-15
Include constants during enabling

kasimi:
During enabling the ["autoload"](https://github.com/phpbb/customisation-db/blob/3.2.x/composer.json#L70) is not handled which is why the constants need to be included manually. This should fix the error reported here: https://www.phpbb.com/community/viewtopic.php?p=14613446#p14613446


VSEphpbb:
You are supposed to run composer before installing this extension.

PR #202, opened 2017-01-15
Move defined constants to class constants

kasimi:
All constants define()d in `includes/constants.php` were moved to class constants in `ext.php`. The `includes/constants.php` is still there with a `deprecated` message at the top.

I also reverted the commit from earlier (1ed3b21) as it still caused errors when deleting data.

### DIFF
--- a/attachment/attachment.php
+++ b/attachment/attachment.php
@@ -14,6 +14,7 @@
 namespace phpbb\titania\attachment;
 
 use phpbb\titania\access;
+use phpbb\titania\ext;
 
 class attachment extends \phpbb\titania\entity\database_base
 {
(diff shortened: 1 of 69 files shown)


VSEphpbb:
Since you changed the composer.json file, the composer.lock file will need to be updated here too:

run: `composer update`

kasimi:
Correct, but the `constants.php` file is nowhere mentioned in the `composer.lock` so it doesn't change, unless I messed something up?

VSEphpbb:
Well then I guess it's OK

DavidIQ:
This pr has conflicts now. Please resolve so they can be merged.

kasimi:
Should be ready to be merged now.

PR #216, opened 2017-05-30
Allow download button to grow in height if needed

VSEphpbb:
Turns this malarkey:
screen shot 2017-05-29 at 5 45 35 pm

Into this:
screen shot 2017-05-29 at 5 45 52 pm


DavidIQ:
I think that malarkey should be un-malarkeyd. Why do we show all versions the contribution applies to like that? Looks terrible...

VSEphpbb:
I assume someone wanted it to show it supports all those versions. We should still have this change regardless.

PR #224, opened 2017-08-01
Use a newer screenshot image viewer

VSEphpbb:
Also fixes https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-730


styles/prosilver/template/contributions/contribution_header.html
```diff
@@ -1,16 +1,5 @@
 <!-- INCLUDE overall_header.html -->
 
-<!-- IF U_DOWNLOAD and U_SUBSCRIBE and not IS_SUBSCRIBED and not QUEUE_STATUS -->
```
DavidIQ: Why was this removed?

styles/prosilver/theme/colorbox.css
```diff
@@ -1,70 +0,0 @@
-/*
```
DavidIQ: I assume this file was just deleted and github's GUI is being dumb and simply showing all lines were removed?

DavidIQ:
If we're no longer needing colorbox then maybe `assets/javascript/jquery.colorbox-min.js` should also be removed and the reference to it removed from `styles/all/template/event/overall_footer_after.html`?

VSEphpbb:
@DavidIQ `assets/javascript/jquery.colorbox-min.js` is being removed along with any references to it.

styles/prosilver/theme/colorbox.css
VSEphpbb: yep

styles/prosilver/template/contributions/contribution_header.html
VSEphpbb: Because this is essentially dead code. From what I have tried to infer, this appeared to be part of an option to Subscribe to a contribution when you have chosen to download one. There is no more code in Titania that unhides this...nor I have I ever remembered experiencing such a thing.

VSEphpbb:
FYI the plugin can be demo'ed here
https://noelboss.github.io/featherlight/gallery.html

DavidIQ:
Yeah, magnific is one I use in ReIMG so... 👍

PR #225, opened 2017-08-01
Fix issues with outdated titania routes

VSEphpbb:
Note EPV can not be updated until after this PR is merged. When EPV is updated it will install Symfony 3 components that will break things because of the deprecated YAML code that this PR fixes.


DavidIQ:
Is this compatible with the current Symfony being used?

VSEphpbb:
Yes it's fully BC

PR #227, opened 2017-08-01
Update to latest dev version of EPV

VSEphpbb:
@paul999 Can we do this? Only v0.0.8 of EPV seemed to be getting installed. This PR updates the lock to the latest dev-master of EPV. It makes me wonder if or how the latest changes to EPV are even running on .com yet?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,14 +52,14 @@
         }
     ],
     "require": {
-    	"php": ">=5.3.3",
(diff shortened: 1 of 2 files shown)


paul999:
I didn't release it yet because I want it tested first; before actually bringing it into production.

VSEphpbb:
OK, closing for now. I'll leave EPV implemting up to you @paul999

PR #238, opened 2017-09-14
Repack diff

kasimi:
So Titania comes with a diff class that makes use of phpBB's core files for diffing that was written in 2008 but was never used. I updated it only where necessary and didn't overhaul it completely.

Notes:
* Previously, each entry in a diff result had a header along the lines of:
    ```
    Index: file.php
    =====================================
    ```
    I removed this.
* The resulting diff always included unchanged files. I added a switch to the diff class to be able to configure it to skip equal files, making the diff smaller and more readable.
* Including phpBB's core files for diffing threw an error about redeclaring classes when using Titania's `_include()` method, which is why I added a new optional argument to skip auto-including. I'm not sure if there's a better way to solve this.
* The `code` BBCode didn't always pick up the `diff`syntax because of too much PHP code in it. This is entirely the fault highlight.js, the library used by s9e/highlighter. I added an even listener that allows using `[code lang=xyz]` throughout Titania which forces the given syntax to be used. Thanks to @JoshyPHP for the code.

Comments and testers are welcome!

repack

### DIFF
--- a/language/en/manage.php
+++ b/language/en/manage.php
@@ -126,5 +126,5 @@ $lang = array_merge($lang, array(
 	'VALIDATION_PV'				=> 'Prevalidator Notes',
 	'VALIDATION_QUEUE'			=> 'Validation Queue',
 	'VALIDATION_SUBMISSION'		=> 'Validation Submission',
-    'VALIDATION_DIFF'			=> 'Repack Diff',
+    'VALIDATION_REPACK_DIFF'	=> 'Repack Diff',
 ));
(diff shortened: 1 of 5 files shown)


language/en/manage.php
```diff
@@ -126,4 +126,5 @@
 	'VALIDATION_PV'				=> 'Prevalidator Notes',
 	'VALIDATION_QUEUE'			=> 'Validation Queue',
 	'VALIDATION_SUBMISSION'		=> 'Validation Submission',
+    'VALIDATION_DIFF'			=> 'Repack Diff',
```
DavidIQ: I think this should probably be `VALIDATION_REPACK_DIFF`.

paul999:
I wonder if we should introduce the same for any revision that is submitted

kasimi:
Those diffs could get quite big though, especially with updated dependencies in the `vendor` folder. Not sure if they are handy enough in all situations.

vinny:
@kasimi @DavidIQ 

Should this only work for extensions? Or for all contributions?It is not working with styles.

PR #240, opened 2017-09-22
CLI command for repacking extensions

kasimi:
There are two issues I haven't been able to figure out so far:

* When passing a message with `-m "Message"` the Extensions Robot creates a post in the validation discussion topic. However, its post count isn't increased.

* I tried to fix generating routes as explained [here](http://symfony.com/doc/current/console/request_context.html) but it doesn't seem to work. Specifically, the `["extra"]["version-check"]` object of the `revision_composer_json` value in the database contains wrong data after running the command:
  * `"host"` is `null`
  * `"directory"` has the absolute path on the server prepended: `/32/db/extension/post_numbers` becomes `/vagrant/www/32/db/extension/post_numbers`

Any hints on how to get those issues fixed?

### DIFF
--- a/console/command/extension/repack.php
+++ b/console/command/extension/repack.php
@@ -1,0 +1,404 @@
+<?php
+/**
+ *
+ * This file is part of the phpBB Customisation Database package.
+ *
+ * @copyright (c) phpBB Limited <https://www.phpbb.com>
+ * @license GNU General Public License, version 2 (GPL-2.0)
+ *
+ * For full copyright and license information, please see
+ * the docs/CREDITS.txt file.
+ *
+ */
+
+namespace phpbb\titania\console\command\extension;
+
+use phpbb\config\config;
+use phpbb\db\driver\driver_interface as db;
+use phpbb\language\language;
+use phpbb\routing\router;
+use phpbb\template\template;
+use phpbb\titania\attachment\attachment;
+use phpbb\titania\attachment\operator as attachments;
+use phpbb\titania\config\config as titania_config;
+use phpbb\titania\contribution\extension\type;
+use phpbb\titania\entity\package;
+use phpbb\titania\ext;
+use phpbb\user;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class repack extends \phpbb\console\command\command
+{
+	const COMMAND_NAME = 'titania:extension:repack';
+
+	/** @var language */
+	protected $language;
+
+	/** @var config */
+	protected $config;
+
+	/** @var template */
+	protected $template;
+
+	/** @var db */
+	protected $db;
+
+	/** @var router */
+	protected $router;
+
+	/** @var titania_config */
+	protected $titania_config;
+
+	/** @var type */
+	protected $ext_type;
+
+	/** @var attachments */
+	protected $attachments;
+
+	/** @var string */
+	protected $contribs_table;
+
+	/** @var string */
+	protected $revisions_table;
+
+	/** @var \titania_revision */
+	protected $revision;
+
+	/** @var \titania_contribution */
+	protected $contrib;
+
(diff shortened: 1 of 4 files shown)


language/en/console.php
```diff
@@ -44,4 +44,14 @@
 	'CLI_DESCRIPTION_REBUILD_COMPOSER'		=> 'Rebuild Composer repository',
 	'CLI_REBUILD_COMPOSER_FORCE'			=> 'Force rebuild even if a build is in progress',
 	'CLI_REBUILD_COMPOSER_FROM_FILE'		=> 'Rebuild repository from revision zip files',
+
+    'CLI_DESCRIPTION_EXTENSION_REPACK'      => 'Repack extension revisions',
+    'CLI_EXTENSION_REPACK_EXT_NAME'         => 'vendor/extname to repack. If omitted, all revisions of all extensions are repacked.',
+    'CLI_EXTENSION_REPACK_MESSAGE'          => 'Message to post in each extension’s validation discussion topic.',
```
Elsensee: Here you've used spaces instead of tabs.

console/command/extension/repack.php
```diff
@@ -0,0 +1,404 @@
...
+	 *
+	 * @param titania_config	$titania_config
+	 * @param type				$ext_type
+	 * @param attachments		$attachments
+	 */
+	public function set_titania_dependencies(
+		titania_config $titania_config,
+		type $ext_type,
+		attachments $attachments
+	)
```
Elsensee: I'd personally prefer it to have all the arguments in a single line, actually..

kasimi:
Regarding the first issue (Extensions Robot's post count not increasing): the post is created with type `ext::TITANIA_QUEUE_DISCUSSION` but in the config there's only [`ext::TITANIA_SUPPORT`](https://github.com/phpbb/customisation-db/blob/6599973510cf89c42137479faf0ff2d1864fb680/config/config.php#L60) enabled to increase the post count. I don't think it's problem really.

I'm still investigating the wrong host/directory...

DavidIQ:
Yeah not worried about the robot account's post count not going up. This path thing looks familiar though...I'll think about it for a bit.

DavidIQ:
Not coming up with anything. @paul999 do you remember the path issue from before?

paul999:
Nope, can't remember seeing it

VSEphpbb:
kasimi are you testing in a 3.2.x board or 3.3.x/master? You should be testing within 3.2.x.

kasimi:
I'm testing on a 3.2.0 board.

kasimi:
Got this fixed finally. [The route helper creates a new request context on each call](https://github.com/phpbb/phpbb/blob/release-3.2.1/phpBB/phpbb/routing/helper.php#L104-L114) which is why modifying the `router` service had no effect. Setting `force_server_vars` to true was the solution.

Will the arguments have to be moved into a single line? I think it's more readable when they are on separate lines.

console/command/extension/repack.php
```diff
@@ -0,0 +1,382 @@
...
+		$this->template			= $template;
+		$this->db				= $db;
+
+		$this->contribs_table	= TITANIA_CONTRIBS_TABLE;
+		$this->revisions_table	= TITANIA_REVISIONS_TABLE;
+
+		$this->language->add_lang(array('console', 'manage', 'manage_tools', 'contributions'), 'phpbb/titania');
+
+		// As we're in the CLI, we need to force server vars so that the route helper generates correct version check URLs
+		$this->config['force_server_vars'] = true;
```
Nicofuma: I think this will alter the board configuration in database.

Nicofuma:
I have one concern with the $config modification, otherwise looks good to me. And sorry for the delay

kasimi:
[offsetSet()](https://github.com/phpbb/phpbb/blob/release-3.2.1/phpBB/phpbb/config/config.php#L69-L81) changes the value for the current request only. For a persistent change, [set()](https://github.com/phpbb/phpbb/blob/release-3.2.1/phpBB/phpbb/config/db.php#L112-L123) would have to be used.

No worries about the delay!

DavidIQ:
Is that last comment regarding @Nicofuma's board configuration comment in `repack.php`?

kasimi:
Ah, yes.

PR #241, opened 2017-09-24
Only diff whitelisted files when repacking

kasimi:
@vinny reported a problem [in the other PR](https://github.com/phpbb/customisation-db/pull/238#issuecomment-331608863). In his first repack that resulted in an empty diff, a GIF image was deleted from the package. phpBB's differ doesn't skip binary files, and so the diff got corrupted by the null bytes in the GIF header. In the debugger the `$diff` value was truncated, I'm not sure why it ended up being empty in the post.

As a primitive solution for this, I added a fixed set of file extensions that are included in the diff. Files with other extensions are **not** listed in the diff.

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -145,6 +145,14 @@ class config extends \phpbb\titania\entity\base
 			'use_queue'					=> array('default' => true),
 			'allow_self_validation'		=> array('default' => true),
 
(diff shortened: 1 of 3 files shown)


kasimi:
Yes, much better!

PR #242, opened 2017-10-22
Run various checks on extension revision version

kasimi:
These changes make sure that:

- the version property in composer.json is present
- the version property in composer.json and the revision version entered by the user are the same
- the version is stable

The code of the `is_stable_version()` method was taken from [here](https://github.com/phpbb/mpv/blob/master/includes/tests/tests_modx.php#L181-L227) and was found by @paul999 within 17 nanoseconds.

### DIFF
--- a/contribution/extension/type.php
+++ b/contribution/extension/type.php
@@ -349,46 +349,15 @@ class type extends base
 	}
 
 	/**
-	 * Checks if the version is stable (1.2.3) or not (1.2.3-RC1)
+	 * Checks if the version is stable (1.2.3, 4.5.6-PL1) or not (7.8.9-RC1, 0.9.8)
 	 *
 	 * @param string $version
 	 * @return bool
 	 */
 	protected function is_stable_version($version)
 	{
-		$version = strtolower($version);
-
-		$unstable = array(
-			'rc',
-			'alpha',
-			'beta',
-			'dev',
-			'a',
-			'b',
-		);
-
-		$option = implode('|', $unstable);
-
-		// Check the version format and numbering
-		if (preg_match('#((\d+)\.)+(\d+)[a-z]?#', $version, $matches))
-		{
-			if (preg_match('#((\d+)\.)+(\d+)-(' . $option . '+)(\d+{0,)?#i', $version, $matches))
-			{
-				return false;
-			}
-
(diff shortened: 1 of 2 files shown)


contribution/extension/type.php
```diff
@@ -337,6 +349,49 @@ protected function update_phpbb_requirement(array $data)
...
+			'beta',
+			'dev',
+			'a',
+			'b',
+		);
+
+		$option = implode('|', $unstable);
+
+		// Check the version format and numbering
+		if (preg_match('#((\d+)\.)+(\d+)[a-z]?#', $version, $matches))
```
VSEphpbb: I think we only allow versions like `x.x.x` and `x.x.x-PLx` so it may be better to just test for that, rather than try to guess what unstable versions might be with a fixed group of them (people may invent others). 

Might be able to simplify this whole function down to just:

```php
return (preg_match('#\d\.\d+\.\d+(-pl\d+)?#i', $version) && version_compare('1.0.0', $version) > 0);
```

Even if you don't do this, your existing code does not need `$matches` since you don't use it.

kasimi: You're right, checking for allowed versions is easier. I had to leave the `$matches` in there because `version_compare()` would report a `-PLx` version as being `< 1.0.0`. Only `x.y.z` should be passed to it.

contribution/extension/type.php
```diff
@@ -349,46 +349,15 @@ protected function update_phpbb_requirement(array $data)
...
-		}
-		else
-		{
-			// Not a valid version
-			return false;
-		}
-
-		return true;
+		// version_compare() doesn't understand -PLx
+		return preg_match('#^(\d+\.\d+\.\d+)(-pl\d+)?$#i', $version, $matches) === 1 && version_compare('1.0.0', $matches[1]) >= 0;
```
VSEphpbb: you can just use the $version in the compare function.

 If the preg match fails, the version compare will; not be called and false will be returned.

Even better would be to use phpbb_version_compare which is case insensitive so you don't need to use strtolower.

kasimi: I overlooked the fact that `version_compare()` only handles `-pl` in lower case. Should be good now. 😄

PR #247, opened 2018-09-03
Auto-select previously selected phpBB branches

battye:
Look at the most recent revision for a contribution to find the previously selected branches, and then auto-select them for a new revision.

Don't try and auto-select anything if it's the first revision, eg. the condition `if (sizeof($this->contrib->revisions))` must be satisfied.

https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-737

### DIFF
--- a/controller/contribution/revision.php
+++ b/controller/contribution/revision.php
@@ -209,6 +209,22 @@ class revision extends base
 	{
 		$this->setup($contrib_type, $contrib);
 
+		$settings = array();
+
+        // Load the revisions for this contribution
+        $this->contrib->get_revisions();
+
+		if (sizeof($this->contrib->revisions))
+        {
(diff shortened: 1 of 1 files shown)


DavidIQ:
Looks good. Just a question: what happens when a previously selected branch is no longer available?

battye:
This change just modifies the pre-selected branches which get passed into `generate_phpbb_version_select()` as the first argument at:

https://github.com/phpbb/customisation-db/blob/b7c58c3e8899d587b6f1799f47c925f63fafb059/display.php#L432

In there we just loop over `$branches` so if one of the previously selected branches has been removed, it won't show as the `(is_array($selected) && in_array($branch, $selected))` won't be satisfied.

DavidIQ:
Perfect, thanks.

PR #249, opened 2018-09-12
Remove hardcoded repository links

VSEphpbb:
This would need to be tested on phpBB.com.

Go to `Manage > Administration > Rebuild Composer repository`

Then check the files created in `ext > phpbb > titania > composer_packages > prod > packages-extension-1.json`

Verify that the extensions listed in them have correct contrib and download URLs, e.g:

```
"homepage": "https://www.phpbb.com/customise/db/extension/ads/",

"dist": {
     "url": "https://www.phpbb.com/customise/db/download/156996/composer",
     "type": "zip"
 }
```
Unfortunately, I have no idea why @prototech hard coded this, but it should not be needed.

### DIFF
--- a/manage/tool/composer/rebuild_repo.php
+++ b/manage/tool/composer/rebuild_repo.php
@@ -221,7 +221,7 @@ class rebuild_repo extends base
 					$group_count = $group = 1;
 				}
 				$last_type = $revision['contrib_type'];
-				/*$download_url = $this->path_helper->strip_url_params(
+				$download_url = $this->path_helper->strip_url_params(
 					$this->controller_helper->route('phpbb.titania.download',
 						array(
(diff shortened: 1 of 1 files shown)


paul999:
This can't be tested locally? I don't like testing on .com...

DavidIQ:
This should be testable locally. Not really sure if maybe the `path_helper` class was just buggy and it didn't work when this was initially done so would be a good idea to ensure it at least does what it's supposed to locally.

VSEphpbb:
I did test locally and it works. The question for me was whether this was specific to .com so that can only truly be answered after deployment. 

It should be fine, anyways phpbb.com is not using these composer repositories yet (that's something intended for phpBB 3.3 extension installation). So deploying and testing should be OK. Worst case we roll it back.

PR #256, opened 2018-09-14
CUSTDB-776 - Remove hooks system

VSEphpbb:
https://tracker.phpbb.com/browse/CUSTDB-776

Regarding changes needed on dot-com as result of this PR:

~~This migration will add two new columns, to the _posts and _topics titania tables. However we are already using columns for the same purpose on dot-com, but the names are bad. So someone will have to go into our DB and delete the newly added columns, and rename the existing old one's with the new names as follows:
change `cdb_posts`table, `phpbb_post_id` field to `post_queue_id`
change `cdb_topics`table, `phpbb_topic_id` field to `topic_queue_id`~~

Then we also need to update the dot-com's custom config settings file with the IDs of the validation queue forums as seen [here](https://github.com/phpbb/customisation-db/compare/3.2.x...VSEphpbb:unhooked?expand=1#diff-215adb4f17e6c57786bec362c538ccb8) with these values:
```
'titania_extensions_queue'    => array(
    ext::TITANIA_QUEUE_DISCUSSION    => 516
    ext::TITANIA_QUEUE               => 511
    'trash'                          => 521
),
'titania_mods_queue'        => array(
    ext::TITANIA_QUEUE_DISCUSSION    => 61
    ext::TITANIA_QUEUE               => 38
    'trash'                          => 28
),
'titania_styles_queue'        => array(
    ext::TITANIA_QUEUE_DISCUSSION    => 87
    ext::TITANIA_QUEUE               => 40
    'trash'                          => 83
),
```
This will be handled here: #257


paul999:
Please, no renaming or deleting of those columns. These tables are way to big and this might cause unneeded downtime to .com.

DavidIQ:
Alright then just leave the columns with the same name. Those tables are very likely very large indeed.

VSEphpbb:
Updated

DavidIQ:
Just 2 questions on this:

1. What happens in the migration if the columns already exist? Does it just skip trying to add them?
2. Has @battye or someone else tested? As long as all queue functions still work and properly publish to the forum then I think this is good to go.

VSEphpbb:
1. the migration is skipped if the column phpbb_post_id is already there.
2. i tested and appears to work, but someone else should still test it too. This is pretty much entirely the stuff that handles creating posts in forums set up as validation queue discussions, like 
https://www.phpbb.com/community/viewforum.php?f=5
So when a contrib is created or new revision submitted, a validation topic is also created. And replies in the queue are added to the forum discussion, and when contribs are approved or denied or whatever, their topics are moved to a trash discussion forum.

I didn't change how any of these functions work, just moved them from hooks into the core, and made sure they still had access to the variables they use.

DavidIQ:
Ok sounds reasonable. @battye when you have a moment please have another look at this and see about doing a smoke test locally.

battye:
Not a problem, once I've finished #260 I'll get on this.

DavidIQ:
For the error it kind of looks like we're trying to insert a post notification for the delete, which doesn't sound right.

battye:
There's a recursive function which is being a bit of a pain to debug, but it seems to be happening in `delete_notifications()`

The stacktrace is like:

delete_notifications() - notification/manager
delete_posts() - functions_admin
delete_post() - functions_posting
forum_queue_hard_delete()

battye:
I just spent 20 minutes trying to replicate the hard delete issue again but it's not happening, so I think we can chalk it up to a dodgy caching issue (I cleared cache before I started again) on my local environment. So I think it's all good! Functionality all seems intact and I haven't come across anything else which would be a roadblock.

config/config.php
```diff
@@ -121,6 +121,23 @@ public function set_default_config()
 				'32'	=> 0,
 			)),
 
+			// Extension/style validation queue forums
+			'titania_extensions_queue'	=> array('default' => array(
+				ext::TITANIA_QUEUE_DISCUSSION	=> 0,
+				ext::TITANIA_QUEUE				=> 0,
+				'trash'							=> 0,
```
DavidIQ: Maybe we need a constant for this too? Seems strange to have constants for everything but this.

PR #257, opened 2018-09-14
Add validation queue forums to dot-com config settings

VSEphpbb:
This depends on #256

### DIFF
--- a/config.php
+++ b/config.php
@@ -39,6 +39,25 @@ $config = array(
 	),
 
 	/**
+	* IDs of forums used for extension/styles validation queue topics
+	*/
+	'titania_extensions_queue' => array(
+		ext::TITANIA_QUEUE_DISCUSSION => 516,
+		ext::TITANIA_QUEUE            => 511,
+		'trash'                       => 521,
+	),
+	'titania_mods_queue'       => array(
+		ext::TITANIA_QUEUE_DISCUSSION => 61,
+		ext::TITANIA_QUEUE            => 38,
+		'trash'                       => 28,
+	),
+	'titania_styles_queue'     => array(
+		ext::TITANIA_QUEUE_DISCUSSION => 87,
+		ext::TITANIA_QUEUE            => 40,
+		'trash'                       => 83,
+	),
+
+	/**
 	* IDs of account used for topic/post release in database forum
 	*/
 	'forum_extension_robot' => 77503,


battye:
This might not be relevant, but as part of the #256 testing I updated my local config and for mine to work I had to use this format with the `array('default' => array( ...` otherwise I was getting issues like `$forum_id` being set to 0, or errors like `Unknown property: titania_extensions_queue`, etc.

````php
			'titania_extensions_queue'    => array('default' => array(
				ext::TITANIA_QUEUE_DISCUSSION    => 4, // id of Forum (Extension Queue Discussion)
				ext::TITANIA_QUEUE               => 3, // id of Forum Category (Extension Validation Queue)
				'trash'                          => 5, // id of Forum (Extension Trash Can)
			)),


			// Accounts to use for posting in the forum
                        ...
			'forum_extension_robot'		=> array('default' => 50),
````

**Comment by VSEphpbb** [comment]
You modded the config/config.php file built in to titania instead of using the extra config-example file? This is a dot-com specific file version of the config-example.php and its settings are not as multi dimensional, and is at the root level.

**Comment by battye** [comment]
Yeah, I'm using ` titania/config/config.php`.

    phpbb.titania.config:
        class: phpbb\titania\config\config
        arguments:
            - '@config'
            - '%phpbb.titania.root_path%'
            - '%core.php_ext%'

Tracing back to services.yml and titania/common.php it looks like titania/config/config.php is used everywhere for me. How are you getting the config.example.php (renamed to config.php in the root folder I guess?) to become initialised instead of the `phpbb\titania\config` version?

**Comment by VSEphpbb** [comment]
I'm pretty sure when I tested I just hacked the same file you did.

But the example config settings file is brought in here: https://github.com/phpbb/customisation-db/blob/3.2.x/config/config.php#L43

File: config.php [hunk 1]
```diff
@@ -38,6 +38,25 @@
 		'32'	=> 616,
 	),
 
+	/**
+	* IDs of forums used for extension/styles validation queue topics
+	*/
+	'titania_extensions_queue' => array(
+		ext::TITANIA_QUEUE_DISCUSSION => 516,
+		ext::TITANIA_QUEUE            => 511,
+		'trash'                       => 521,
```
DavidIQ: Maybe we need a constant for this too? Seems strange to have constants for everything but this.

VSEphpbb: All the constants in this extension seem to be just for magic numbers.

PR #259, opened 2018-09-15
Fix for contribs in hidden categories still showing everywhere

battye:
From https://tracker.phpbb.com/browse/CUSTDB-711

I wasn't able to replicate the first dot point, but I was for the second. Presumably there are reasons why someone might want to make a category invisible without totally cutting off access to contribs inside, so what I've done is made it so that the contribs from the hidden categories won't show on the "all" listing or on the parent category listing either. So the only way you can view it is if you explicitly go into that hidden category (eg. maybe someone wants to keep it private and supplies a direct link to trusted people or something).

How it looks now:

kv4

### DIFF
--- a/includes/overlords/contribs.php
+++ b/includes/overlords/contribs.php
@@ -220,7 +220,7 @@ class contribs_overlord
 				);
 
 				// Simplify it - just make it an array
-				if (is_integer($hierarchy_ids))
+				if (!is_array($hierarchy_ids))
 				{
 					$hierarchy_ids = array($hierarchy_ids);
 				}
@@ -379,7 +379,7 @@ class contribs_overlord
 			{
 				$contrib_categories = explode(',', $row['contrib_categories']);
 
-				if (count(array_diff($contrib_categories, $hidden_categories_ids)) == 0)
+				if (!count(array_diff($contrib_categories, $hidden_categories_ids)))
 				{
 					// Don't show it on the "all" listings page, because the contrib is only in hidden categories
 					continue;
--- a/includes/overlords/contribs.php
+++ b/includes/overlords/contribs.php
@@ -108,6 +108,34 @@ class contribs_overlord
 	}
 
 	/**
+	 * Get list of the hidden categories
+	 * @return array
+	 */
+	public static function get_hidden_categories()
+	{
+		$hidden_categories_ary = array(
+			'SELECT'	=> 'c.category_id',
+
+			'FROM'		=> array(
+				TITANIA_CATEGORIES_TABLE	=> 'c',
+			),
+
+			'WHERE'		=> 'c.category_visible = 0',
+		);
+
+		$hidden_categories_sql = phpbb::$db->sql_build_query('SELECT', $hidden_categories_ary);
+		$hidden_categories_result = phpbb::$db->sql_query($hidden_categories_sql);
+		$hidden_categories_ids = array();
+
+		while ($hidden_categories_row = phpbb::$db->sql_fetchrow($hidden_categories_result))
+		{
+			$hidden_categories_ids[] = (int)$hidden_categories_row['category_id'];
+		}
+
+		return $hidden_categories_ids;
+	}
+
+	/**
 	 * Display contributions
 	 *
 	 * @param string $mode The mode (category, author)
@@ -124,6 +152,7 @@ class contribs_overlord
 
 		$tracking = phpbb::$container->get('phpbb.titania.tracking');
 		$types = phpbb::$container->get('phpbb.titania.contribution.type.collection');
+		$cache = phpbb::$container->get('phpbb.titania.cache');
 
 		// Setup the sort tool if not sent, then request
 		if ($sort === false)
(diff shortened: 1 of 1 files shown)


includes/overlords/contribs.php
```diff
@@ -290,8 +362,21 @@ public static function display_contribs($mode, $id, $branch = false, $sort = fal
...
 
 		$contrib_ids = $user_ids = array();
+
 		while ($row = phpbb::$db->sql_fetchrow($result))
 		{
+			if ($check_hidden_categories_on_all)
+			{
+				$contrib_categories = explode(',', $row['contrib_categories']);
+
+				if (sizeof($contrib_categories) == 1 && in_array($contrib_categories[0], $hidden_categories_ids))
```
VSEphpbb: Please use `count` instead of `sizeof`and also `===`

includes/overlords/contribs.php
```diff
@@ -175,6 +207,43 @@ public static function display_contribs($mode, $id, $branch = false, $sort = fal
...
+						{
+							$all_subcategory_ids[] = (int) $child_id;
+						}
+					}
+				}
+
+				// We'll only remove categories if these are not equal. If they ARE equal then it means we are currently
+				// inside the hidden category (as the other ids will be the subcategories)
+				asort($all_subcategory_ids);
+				asort($id);
```
VSEphpbb: Please fix this function's docblock which says `$id` is an integer. Obviously it can be an array of ids.

includes/overlords/contribs.php
```diff
@@ -175,6 +207,43 @@ public static function display_contribs($mode, $id, $branch = false, $sort = fal
...
+				// Get a list of hidden categories, so that if a category has been set to invisible we won't show
+				// the contribs in the parent category listing
+				$all_subcategory_ids = array();
+
+				foreach ($hidden_categories_ids as $hidden_category_id)
+				{
+					$all_subcategory_ids[] = (int) $hidden_category_id;
+					$children_ids = array_keys($cache->get_category_children($hidden_category_id));
+
+					foreach ($children_ids as $child_id)
```
VSEphpbb: So this will only check one level of children, and presumably miss further potential nested layers of children?

includes/overlords/contribs.php
```diff
@@ -200,7 +269,9 @@ public static function display_contribs($mode, $id, $branch = false, $sort = fal
 						)
 					),
 
-					'WHERE'		=> ((is_array($id) && sizeof($id)) ? phpbb::$db->sql_in_set('cic.category_id', array_map('intval', $id)) : 'cic.category_id = ' . (int) $id) . '
+					// If multiple categories, use the stripped list. If it's just the single hidden category, that's okay
+					// as presumably someone has gone looking for the hidden category, or it has been linked to, etc.
+					'WHERE'		=> ((is_array($visible_category_ids) && sizeof($visible_category_ids)) ? phpbb::$db->sql_in_set('cic.category_id', array_map('intval', $visible_category_ids)) : 'cic.category_id = ' . (int) $id) . '
```
VSEphpbb: Please use `count` instead of `sizeof`. Also, now `$id` is an array but being cast to an integer? How does that work?

Seems to me like $id can be either an int or an array so that needs to be checked above, before we start treating it as an array by default. And also it needs to be checked here like it was in the old code before we assume it is an integer.

includes/overlords/contribs.php
battye: Why do you want to start using `count`? I can't find any references to `count` in all of Titania outside of dependencies. Everything that's currently there uses `sizeof` (probably because that was the standard for 3.0.x).

battye: Okay I asked @marc1706 - we're using `count` in phpBB 3.2 now. I'll use `count` here as well given that information, but I still think it's weird that we'd introduce the inconsistency into Titania.

includes/overlords/contribs.php
```diff
@@ -107,6 +107,34 @@ public static function get_contrib_object($contrib_id, $query = false)
...
+
+			'FROM'		=> array(
+				TITANIA_CATEGORIES_TABLE	=> 'c',
+			),
+
+			'WHERE'		=> 'c.category_visible = 0',
+		);
+
+		$hidden_categories_sql = phpbb::$db->sql_build_query('SELECT', $hidden_categories_ary);
+		$hidden_categories_result = phpbb::$db->sql_query($hidden_categories_sql);
```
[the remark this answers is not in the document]
battye: Done

includes/overlords/contribs.php
```diff
@@ -175,6 +207,43 @@ public static function display_contribs($mode, $id, $branch = false, $sort = fal
...
+						}
+					}
+				}
+
+				// We'll only remove categories if these are not equal. If they ARE equal then it means we are currently
+				// inside the hidden category (as the other ids will be the subcategories)
+				asort($all_subcategory_ids);
+				asort($id);
+
+				if (serialize(array_values($all_subcategory_ids)) != serialize(array_values($id)))
```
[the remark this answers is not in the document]
battye: Done

includes/overlords/contribs.php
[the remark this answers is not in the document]
battye: Done

includes/overlords/contribs.php
battye: Updated

includes/overlords/contribs.php
battye: I've rewritten this and the other one

battye:
@kasimi and @VSEphpbb I've made a bunch of changes to this

PR #260, opened 2018-09-16
Restore automated validation for 3.2 translations

battye:
Display a message to the user if they try and upload a language pack that doesn't contain a LICENSE file.

> You must include a 'LICENSE' file in the translation package. 

kv5

@Crizz0: Can a language pack contain more than one language? Like the example you gave me just had `\language\de` in it. But if there can be more than one language in that folder then I'll need to tweak this change.

Actually, come to think if it I might change this anyway do scan `language/` for directories and check to see if LICENSE is in each subdirectory, because the way it's currently written a user could whack a LICENSE file anywhere and this would validate. I guess it's best to make sure they've put in the language root.

https://tracker.phpbb.com/browse/CUSTDB-778

### DIFF
--- a/contribution/translation/prevalidator.php
+++ b/contribution/translation/prevalidator.php
@@ -44,14 +44,23 @@ class prevalidator
 		'language/en/LICENSE',
 		'language/en/CHANGELOG',
 		'language/en/VERSION',
-
-		// Ignore markdown files too
-		'language/en/AUTHORS.md',
-		'language/en/README.md',
-		'language/en/LICENSE.md',
-		'language/en/CHANGELOG.md',
-		'language/en/VERSION.md',
 	);
+
+	/**
+	 * File extensions list
+	 * @var array
+	 */
+	protected $ignore_files_extensions = array(
+		'',
+		'.md',
+		'.txt'
+	);
+
+	/**
+	 * Computed ignore files
+	 * @var array
+	 */
+	private $ignore_listing = array();
 
 	/**
 	 * For some reason we don't put index.htm in these directories
@@ -76,6 +85,17 @@ class prevalidator
 	}
 
 	/**
+	 * Check to see if a file should be ignored
+	 * @param $file
+	 * @return bool
+	 */
+	private function is_in_ignored_list($file)
+	{
+		// Is the specified file in our listing?
+		return in_array(strtolower($file), $this->ignore_listing);
+	}
+
+	/**
 	 * Get helper.
 	 *
 	 * @return \phpbb\titania\contribution\prevalidator_helper
@@ -97,6 +117,19 @@ class prevalidator
 	{
 		$package->ensure_extracted();
 		$error = $missing_keys = array();
+
+		// Build up the list of files to ignore; do it here because we only want to run this computation once
+		if (!count($this->ignore_listing))
+		{
+			foreach ($this->ignore_files as $ignore_file)
+			{
+				foreach ($this->ignore_files_extensions as $ignore_files_extension)
+				{
+					// Build up a list of our ignored files with the different extensions
+					$this->ignore_listing[] = strtolower($ignore_file . $ignore_files_extension);
+				}
+			}
+		}
 
 		// Basically the individual parts of the translation, we check them separately, because they have colliding filenames
 		$types = array(
@@ -258,7 +291,7 @@ class prevalidator
 		foreach ($list_uploaded_files as $file)
 		{
 			$in_reference_list = in_array($file, $list_reference_files);
-			$in_ignore_list = in_array($file, $this->ignore_files);
+			$in_ignore_list = $this->is_in_ignored_list($file);
 
 			if (!$in_reference_list && !$in_ignore_list)
 			{
@@ -417,7 +450,7 @@ class prevalidator
 		$dp = opendir($root_dir . $dir);
 		while (($fname = readdir($dp)))
 		{
-			if (is_file("$root_dir$dir$fname") && !in_array(strtoupper($fname), $this->ignore_files))
+			if (is_file("$root_dir$dir$fname") && !$this->is_in_ignored_list($fname))
 			{
 				$matches[$dir][] = $fname;
 			}
--- a/contribution/translation/prevalidator.php
+++ b/contribution/translation/prevalidator.php
@@ -13,6 +13,8 @@
 
 namespace phpbb\titania\contribution\translation;
 
+use Symfony\Component\Finder\Finder;
+
 /**
  * Translation prevalidator
  *
@@ -42,6 +44,13 @@ class prevalidator
 		'language/en/LICENSE',
 		'language/en/CHANGELOG',
 		'language/en/VERSION',
+
+		// Ignore markdown files too
+		'language/en/AUTHORS.md',
+		'language/en/README.md',
+		'language/en/LICENSE.md',
+		'language/en/CHANGELOG.md',
+		'language/en/VERSION.md',
 	);
 
 	/**
@@ -81,10 +90,12 @@ class prevalidator
 
 		// Basically the individual parts of the translation, we check them separately, because they have colliding filenames
 		$types = array(
+			'ext' => 'ext/phpbb/viglink/language/en/',
 			'language' => 'language/',
-			'prosilver' => 'styles/prosilver/imageset/en',
-			'subsilver2' => 'styles/subsilver2/imageset/en',
+			'prosilver' => 'styles/prosilver/theme/en/',
 		);
+
+		$iso_code = '';
 
 		// Do the check for all types
 		foreach ($types as $type => $path)
@@ -99,13 +110,16 @@ class prevalidator
 			{
 				case 'language':
 					// The uploaded files array has keys prefixed with the upload path of the contribution
-					// Have it stored in the variable so we can work with it
-					$uploaded_files_prefix = explode('/', key($uploaded_files));
+					// Have it stored in the variable so we can work with it. Need to get the first key to ignore viglink
+					$uploaded_files_keys = array_keys($uploaded_files);
+
+					$uploaded_files_prefix = explode('/', $uploaded_files_keys[1]);
 					$iso_code = $uploaded_files_prefix[2];
 					$uploaded_files_prefix = $package->get_temp_path() . '/' . $uploaded_files_prefix[0];
 
 					// This goes directly to the root of the uploaded language pack, like /upload_path/language/cs/
-					$uploaded_lang_root = $uploaded_files_prefix . '/language/' . $iso_code . '/';
+					$uploaded_language = $uploaded_files_prefix . '/language/';
+					$uploaded_lang_root = $uploaded_language . $iso_code . '/';
 
 					// Just perform a basic check if the common file is there
 					if (!is_file($uploaded_lang_root . 'common.php'))
@@ -141,10 +155,6 @@ class prevalidator
 								{
 									$missing_keys[$dir . $file] = $this->check_missing_keys($reference_filepath . '' .	$dir . $file, $uploaded_file_path);
 								}
-
-								// In the last step we have removed the license and index files if there were any. We'll just put a new one instead
-								$this->add_license_files($uploaded_lang_root . '/LICENSE', $reference_filepath);
-								$this->add_htm_files($uploaded_lang_root, $reference_filepath);
 							}
 						}
 					}
@@ -159,10 +169,15 @@ class prevalidator
 							}
 						}
 					}
+
+					// In the last step we have removed the license and index files if there were any. We'll just put a new one instead
+					$this->add_license_files($uploaded_lang_root . 'LICENSE', $reference_filepath);
+					$this->add_htm_files($uploaded_language, $reference_filepath);
+
 					break;
 
 				case 'prosilver':
-				case 'subsilver2':
+				case 'ext':
 					// just let them go through atm...
 					break;
 			}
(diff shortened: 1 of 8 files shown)


battye:
No need to review this one yet, I've decided to make some changes as outlined above.

Crizz0:
One .zip-archive (package uploaded to the CDB) contains always only one language. E.g. spanish "casual honorifics" (language/es_x_tu/) comes separate from "formal honorifics" (language/es/).

The LICENSE file has to be always in language/{iso}/LICENSE, not in styles/, not in the viglink-language-directory and not in any other subdirectory. :)

battye:
I've been digging more and I found that we did in fact have a translation validator, but it looks like it hasn't been usable since 3.0.x - it has a lot of references to mods and other parts of the directory structure which are no longer relevant.

So I'm in the process of getting that working again. @Crizz0 - the old mechanism was to, regardless of whether the user added a LICENSE file or not, overwrite or create it with whatever the contents of our own LICENSE file in the docs/ directory is.

The other thing it does which is proving to be a bit of a pain, is add index.htm files in a few places. I've tweaked it to add them in subdirectories automatically, but it looks like I'll have to hard code a few directories to exclude. @DavidIQ or @VSEphpbb (or @marc1706) - is there a reason why some directories in phpBB contain index.htm while some don't? For some of these in the example below I'm not sure why we want to restrict the directory listing in the first case but not the second.

> index.htm is in language\en\email **but not language\en\email\short**
> index.htm is in styles\prosilver\theme\images **but not styles\prosilver\theme**
> index.htm is in language\en\ **but not \language\en\help**

Ultimately the translation prevalidator is currently failing because of this, because we check to see whether the files in the translation package matches the phpBB package - but obviously phpBB doesn't have those index.htm files in a few places - so I'll probably have to hard code a fix to this as mentioned above.

> 0 = "The file <code>language/de/email/short/index.htm</code> is not allowed."
> 1 = "The file <code>language/de/help/index.htm</code> is not allowed."
> 2 = "The file <code>styles/prosilver/theme/de/index.htm</code> is not allowed."

battye:
@DavidIQ @VSEphpbb @kasimi If anyone would like the link to my private testing board to try the changes please let me know and I can PM you the link

In summary, I restored the functionality that used to be there but stopped working after 3.0 so things like this will start working again now:

* It'll copy the LICENSE file from phpBB 
* It'll add index.htm files in select places (as above there doesn't seem to be much rhyme or reason to where we do and don't have it now)
* It looks like maybe checks like missing lang keys and missing files may not have been happening? That works again now.

contribution/translation/prevalidator.php
```diff
@@ -42,6 +44,23 @@ class prevalidator
 		'language/en/LICENSE',
 		'language/en/CHANGELOG',
 		'language/en/VERSION',
+
+		// Ignore markdown files too
```
VSEphpbb: instead of adding explicit .md versions, why not just test for the possible existence of any extension to the file names listed above, so we can also catch, for example, .txt as well as .md

contribution/translation/prevalidator.php
```diff
@@ -263,31 +315,85 @@ private function check_missing_keys($reference_file, $uploaded_file)
...
+	/**
+	 * Add license file
+	 * @param $license_file
+	 * @param $reference_path
+	 */
 	private function add_license_files($license_file, $reference_path)
 	{
 		$res = fopen($license_file, 'w');
-		fwrite($res, file_get_contents($reference_path . '/docs/COPYING'));
+		fwrite($res, file_get_contents($reference_path . 'docs/LICENSE.txt'));
```
VSEphpbb: I don't think we use the .txt extension?

kasimi: I assume it's this file? https://github.com/phpbb/customisation-db/blob/3.2.x/docs/LICENSE.txt

contribution/translation/type.php
```diff
@@ -154,10 +154,11 @@ public function translation_validate(\titania_contribution $contrib, \titania_re
 
 		$version = $revision->phpbb_versions[0];
 
-		if ($version['phpbb_version_branch'] != 30)
+		if ($version['phpbb_version_branch'] != 32)
```
kasimi: Is there a constant defined somewhere that we can use here?

contribution/translation/prevalidator.php
```diff
@@ -159,10 +179,15 @@ public function check_package($package, $reference_filepath)
 							}
 						}
 					}
+
+					// In the last step we have removed the license and index files if there were any. We'll just put a new one instead
+					$this->add_license_files($uploaded_lang_root . 'LICENSE', $reference_filepath);
```
kasimi: I'm not familiar with the way the license is checked. What happens if a translation with a wrong license is validated? Is it silently replaced with our copy of the GPLv2 here?

battye: Perhaps @Crizz0 can confirm, but in Poland I recall he said it must always be a copy of the GPL 2 license.

contribution/translation/prevalidator.php
battye: It's the phpBB one: https://github.com/phpbb/phpbb/blob/master/phpBB/docs/LICENSE.txt

The package gets extracted within Titania, so line 326 above would resolve to a path like:

    ./ext/phpbb/titania/store/extracted/3.2.3/docs/LICENSE.txt

contribution/translation/prevalidator.php
battye: I'm happy to make that change, but maybe we should restrict it purely to no extension, .md and .txt? That way we'll get a validation error if someone put garbage like `LICENSE.exe` or something.

contribution/translation/type.php
battye: I can't find anything. We could take the highest version number from https://github.com/phpbb/customisation-db/blob/3.2.x/config/config.php#L158 maybe?

kasimi: Sounds good!

If we want to plan ahead, maybe we could add a new config entry in that file. At some point we might want to be able to allow language pack submissions for 3.2.x and 3.3.x at the same time.

battye:
@VSE I've added a section where we can specify the allowed file extensions and then create the ignored files list dynamically and @kasimi for now I've just got it checking against `phpbb_versions` because we've already got the `allow_uploads` attribute on there to match against But yeah it's easy enough to add a new config section for translations if we want to handle that separately.

	'phpbb_versions' => array('default' => array(
		'20'	=> array( ... 'allow_uploads' => false),
		'30'	=> array( ... 'allow_uploads' => false),
		'31'	=> array( ... 'allow_uploads' => false),
		'32'	=> array( ... 'allow_uploads' => true),
	)),

Crizz0:
Is there a way, I could test this with a few recent uploaded language packages before mergen into the cdb?

battye:
> Is there a way, I could test this with a few recent uploaded language packages before mergen into the cdb?

Yes I can arrange that, I can give you access to my private test environment. Will you be on IRC in the next 8 hours or so?

Crizz0:
No. I have breakfast now and go to work in any minute.

Crizz0:
I tested it a bit and I would like to see a "submit anyway option", when the test failed. Something like "The tests failed, due to following reason. Do you want to upload the packages anyway?"

Due to some languages who need some extras, which cannot always be considered in the recent version of validator, I would still like to have this decision in the hands of the "Translations Manager".

battye:
I don't think that will be difficult to add. 👍 
So something like that message you suggested above (if there were failures) with a checkbox below with a "Submit with validation errors" label?

battye:
Do you mind giving it another test @Crizz0? I've made some changes. The only thing which caught my eye is it seems to be asking whether you want to proceed twice. I'm not sure if that's because of my change or whether it was like that before. I have a feeling it was like that before because if you try uploading a language pack without a LICENSE file (like the original question :grin:) then on one of the stages it repacks and says `Your language pack has been repacked and has passed the validation process which checks for missing keys, structure, additional files and license. Please continue.` which seems like a valid reason to have that second step.

controller/contribution/revision.php
```diff
@@ -745,7 +751,20 @@ protected function process_steps()
 
 		$hash = $this->package->get_md5_hash();
 
-		$result = $this->run_step($step['function']);
+		$ignore_validation_errors = $this->request->variable('ignore_validation_errors', 0);
+
+		if ($ignore_validation_errors && $step['function'][1] == 'translation_validate')
```
battye: @VSE or @DavidIQ - I really dislike this line as it feels like such a dirty hack. Do you know if there's a more elegant way to check that the type of revision we are dealing with at this point is a translation?

VSEphpbb: I don't know, but "elegant" is not a word I would ever associate with Titania's code 😉

DavidIQ: I think this is fine, but will need to make sure that submitting extensions and styles and other contributions aren't broken as I'm not sure what happens when that function key isn't present or isn't an array.

battye:
Updated

styles/prosilver/template/contributions/contribution_revision.html
```diff
@@ -164,7 +164,7 @@ <h3>{L_AGREEMENT}</h3>
 				{AUTOMOD_RESULTS}
 			<!-- ENDIF -->
 			<!-- IF S_PASSED_TRANSLATION_VALIDATION -->
-				{L_PASSED_VALIDATION}
+				<div style="text-align:center;">{L_PASSED_VALIDATION}</div>
```
Crizz0: Do we not have proper css for that?

battye: We do now! :smile:

DavidIQ:
I totally missed mentioning this and thought it was mentioned in the ticket, so sorry: https://github.com/phpbb/phpbb-translation-validator

That repository has the translation validation logic and is a tool the translators use on its own. I think that perhaps this tool should be used/integrated instead of having separate logic within Titania? Not sure what @Crizz0 thinks about this.

Crizz0:
I thought this extension uses the validator   and did not create it's own version, which will double the maintenance effort.

battye:
> I thought this extension uses the validator and did not create it's own version, which will double the maintenance effort.

These changes just restores functionality that used to exist in Titania and got lost along the way.

DavidIQ:
@battye can you research if it would be possible to use the existing language pack validator for this instead of having separate logic to maintain in Titania? If needed we could maybe just convert it to an "LPV" (Language Pre Validator) and have that available.

battye:
I'll have a look

battye:
@DavidIQ @Crizz0 

The [EPV](https://www.phpbb.com/extensions/epv/) runs on the site using PHPUnit tests, so how we've done that I'm sure we could repeat for the Translation Validator - I guess the obstacles are the same whether it's in Titania or just a page on the website (I think I would opt for integration in Titania).

The big roadblock seems to be how it could be integrated without actually including the PHP files as prototech mentioned above. It's something that happens a few times throughout the Translation Validator: https://github.com/phpbb/phpbb-translation-validator/search?q=include&unscoped_q=include

Perhaps Crizzo can confirm if there is anything else it is required for, but from what I can see the only reason lang files are directly included (and what would preclude us from running it on .com) is to match the language keys between the language pack and the default phpBB language keys (make sure what should be there is there, make sure that nothing is missing, etc).

So what I've been playing around with is a way to parse the language files rather than just include them. I found [a script](http://jsteemann.github.io/blog/2015/06/16/parsing-php-arrays-with-php/) where someone has attempted to parse a string containing a PHP array, and I've built on that to include allowing multi-level of array, square brackets for arrays, multiple array_merge's in the same file, ignoring comments and a few other things which our language file arrays contain. It's a pain but I think it's the only way we can do it? By parsing the text, we can securely get the language keys (and the values if necessary, which in my testing would be fine for everything except `'CLI_HELP_CRON_RUN'` in `cli.php` because it's not just a string literal - it's referencing another language var) and re-create the array manually for comparison.

Here's a few gists I've made to demonstrate what I mean.

* [Language parser script](https://gist.github.com/battye/73aa6d2e93c670f53e609f51e3852550)
* ["de" output](https://gist.github.com/battye/139921712a9f86dccf9f5a5c30f63d76#file-de-txt)
* ["en" output](https://gist.github.com/battye/139921712a9f86dccf9f5a5c30f63d76#file-en-txt)

If you're happy with this idea, I think it'll then be a two part process: 

1. Pretty up and phpbb-ise those classes and then integrate that into the Translation Validator repo. 

2. Providing this doesn't break anything in the T.V., then I can go about integrating the T.V. into Titania somehow (Composer dependency maybe? Will need to look into how we got the EPV and PhpUnit working through the website and replicate that functionality).

DavidIQ:
@battye how about looking into using [ob_get_clean](https://secure.php.net/manual/en/function.ob-get-clean.php) to load the contents into a variable? Something like:

```php
ob_start();
include("common.$phpExt");
$common_lang = ob_get_clean();
```
Might make things easier for parsing through?

Crizz0:
Well, the T.V. checks for the keys, so he compares "en" with any other language. Checks within the keys the variables, html-code etc.

So only it shouldn't be enough to simply load the key-names into an array and check, if it is there.

battye:
Ah okay @Crizz0, it shouldn't be a problem anyway as the values are still parsed.

> @battye how about looking into using [ob_get_clean](https://secure.php.net/manual/en/function.ob-get-clean.php) to load the contents into a variable? Something like:
> 
> ```
> ob_start('common_lang');
> include("common.$phpExt");
> $common_lang = ob_get_clean();
> ```
> 
> Might make things easier for parsing through?

@DavidIQ I don't know if that would help, as the lang files don't actually output anything (unless I misunderstood what you meant?)

paul999:
For security reasons, including any php file on .com within php that is uploaded is a no-go. The problem is that the file is ran, so if a user uploads something with for example exec('rm -rf /');, it will be run ;), there will just be no output.

battye:
This dev work is now continuing over at #278

PR #263, opened 2018-11-23
CUSTDB-780 - Responsive design for queue stats

battye:
https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-780

The queue stats weren't displaying properly for smaller displays. This change makes it look better for four screen sizes:

* Very small (385px width or less)
* Small (385-520)
* Medium (520-890)
* Large (890px and larger)
  * This is the default, the other three settings I added media queries for in `responsive.css`.

The stats graph and legend is centre aligned now, and I've split the approval/denial rate graph into its own section, so it should look like this now:

graph2

### DIFF
--- a/styles/prosilver/theme/responsive.css
+++ b/styles/prosilver/theme/responsive.css
@@ -134,3 +134,4 @@
 		width: 100% !important;
 	}
 }
+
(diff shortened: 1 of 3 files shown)


styles/prosilver/theme/responsive.css
```diff
@@ -77,4 +133,4 @@
 	.contrib-quickview {
 		width: 100% !important;
 	}
-}
+}
```
VSEphpbb: Missing newline at end of file

battye: Added

PR #264, opened 2018-12-06
CUSTDB-765 - Missing index on attachment comment field

battye:
Full discussion at https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-765

I think this might be the reason why the attachment comments become inverted upon saving. The comment_list[] indexes were empty - the template was expecting `ASSOC_INDEX` but that wasn't defined when assigning the template vars.


Galixte:
Thanks!

PR #267, opened 2018-12-12
apple responsive rank img

hanakin:
Pending tests on area51

### DIFF
--- a/styles/prosilver/theme/custom.css
+++ b/styles/prosilver/theme/custom.css
@@ -3,5 +3,11 @@
 	$Custom
 -------------------------------------------------------------- */
 .rank-img img {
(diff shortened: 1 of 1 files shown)


hanakin:
Why was this merged? This will not fix anything this will make everything larger

DavidIQ:
Bigger than it is now _without_ this change???? 😆 

image

You should have closed it then if you knew it was useless. I'll revert.

DavidIQ:
Is there an actual fix coming to correct the abnormality as seen above?

hanakin:
Was not aware this was still an issue on cdb the fix is to remove the media query part for this. The top part is all that is needed which sets them all at 160px

PR #268, opened 2018-12-14
CUSTDB-680 - Filter contributions by status

battye:
New functionality which allows the user to filter by contribution status (all, approved, unapproved).

ch4

https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-680

### DIFF
--- a/assets/javascript/ajax.js
+++ b/assets/javascript/ajax.js
@@ -114,9 +114,10 @@
 			$this.attr('href', res.categories[$this.data('category-id')]);
 		});
 
+		// When the filter options change, make sure the correct option is shown as active on the screen
 		titania.updateSortOptions($('.branch-sort'), $('.branch-sort-options'), res.branches);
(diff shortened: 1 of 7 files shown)


VSEphpbb:
I assume this is only visible to team members?

VSEphpbb:
Non-team members (even unregistred guests) can still see this drop down, and choose approved, unapproved, etc. I think this drop down should only be visible to team members.

DavidIQ:
I missread that part only thinking that the retrieval will not return unapproved items regardless of if the parameter is sent. So yeah the building of the options needs to take into account if the user has queue access to that type. I think not super urgent that needs to be fixed right now since the query to produce the list will exclude by permission anyways.

DavidIQ:
Maybe check how the query is being built to show the contributions list to begin with. There seems to be a differentiation between the general public and a person that has queue access.

PR #270, opened 2018-12-15
Bugfixes for approved status filter

battye:
Bugfixes for https://github.com/phpbb/customisation-db/pull/268

* Don't show the dropdown for types that the user doesn't have validate permissions for
* Add status to the pagination links

### DIFF
--- a/controller/index.php
+++ b/controller/index.php
@@ -347,7 +347,7 @@ class index
 	protected function get_category_type()
 	{
 		$children = $this->get_children_ids();
-		$type_id = (!is_null($this->category)) ? $this->category->category_type : false;
+		$type_id = ($this->category !== null) ? $this->category->category_type : false;
 
 		// If the category is the top most parent, we'll try to get the type from the first child
 		if (!$type_id && !empty($children))
@@ -541,15 +541,14 @@ class index
 	 */
 	private function valid_type_permissions()
 	{
-		$types = \phpbb::$container->get('phpbb.titania.contribution.type.collection');
-		$types_managed = $types->find_authed('validate');
+		$types_managed = $this->types->find_authed('validate');
 
 		// If current type id is null, it's the index page
 		$current_category_type = $this->get_category_type();
-		$current_type_id = ($current_category_type !== false) ? $this->get_category_type()->get_id() : null;
+		$current_type_id = ($current_category_type !== false) ? $current_category_type->get_id() : null;
 
 		// If the user manages some types, and the current type is in that list (or it's the index) show the dropdown.
-		$show = (sizeof($types_managed) && (is_null($current_type_id) || in_array($current_type_id, $types_managed)));
+		$show = (sizeof($types_managed) && ($current_type_id === null || in_array($current_type_id, $types_managed)));
 
 		return $show;
 	}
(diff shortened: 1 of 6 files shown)


controller/index.php
```diff
@@ -529,11 +530,34 @@ protected function assign_status()
...
+	{
+		$types = \phpbb::$container->get('phpbb.titania.contribution.type.collection');
+		$types_managed = $types->find_authed('validate');
+
+		// If current type id is null, it's the index page
+		$current_category_type = $this->get_category_type();
+		$current_type_id = ($current_category_type !== false) ? $this->get_category_type()->get_id() : null;
+
+		// If the user manages some types, and the current type is in that list (or it's the index) show the dropdown.
+		$show = (sizeof($types_managed) && (is_null($current_type_id) || in_array($current_type_id, $types_managed)));
```
VSEphpbb: `is_null` usage again here can be optimized

styles/prosilver/template/index_body.html
```diff
@@ -65,7 +65,7 @@ <h2>{L_CUSTOMISATION_DATABASE}</h2>
 			</div>
 		</div>
 
-		<div class="dropdown-container sort-option status-sort rightside">
+		<div class="dropdown-container sort-option status-sort rightside"<!-- IF not SHOW_STATUS --> style="display: none;"<!-- ENDIF -->>
```
VSEphpbb: Why wouldn't you just wrap this entire HTML block for the drop down in an `<!-- IF SHOW_STATUS -->` condition? Would be better to just not render it to the template at all than use CSS and JS to hide it.

controller/index.php
```diff
@@ -529,11 +530,34 @@ protected function assign_status()
...
+	}
+
+	/**
+	 * Check whether the user has permission to filter by unapproved contributions
+	 * @return bool
+	 * @throws \Exception
+	 */
+	private function valid_type_permissions()
+	{
+		$types = \phpbb::$container->get('phpbb.titania.contribution.type.collection');
```
prototech: Avoid using phpbb::$container if at all possible. The idea generally was to get rid of the phpbb class entirely once the DIC and extension system came into play in 3.2. The type collection is already a dependency of the index controller. Use `$this->types` instead.

controller/index.php
```diff
@@ -529,11 +530,34 @@ protected function assign_status()
...
+		$types_managed = $types->find_authed('validate');
+
+		// If current type id is null, it's the index page
+		$current_category_type = $this->get_category_type();
```
prototech: This variable is useless. The ternary in the following line calls `$this->get_category_type()` again.

sort.php
```diff
@@ -247,6 +251,16 @@ public function get_sort_dir()
 		return $this->sort_dir;
 	}
 
+	/**
+	 * Filter for contribution status
+	 * @return mixed
```
prototech: string

styles/prosilver/template/index_body.html
battye: It needs to be done with JS/CSS because if you click on to a different category it reloads the items with AJAX. 

For example, if we wrapped the entire thing in an `IF` block and the first type the user loaded was a type where the user didn't have validate permissions - if that user then clicked on to another type where they did have permissions, the HTML wouldn't exist to toggle the visibility in JS.

battye:
Thanks @VSEphpbb and @prototech (!! :smiley:) for the reviews, I've pushed those changes.

PR #276, opened 2019-02-11
[CUSTDB-626] Add more information to the download link popup

battye:
This change shows the directory the user should upload the extension to as one of the details when hovering over the download link. (I noticed someone made an error doing this when installing an extension [a few days ago](https://www.phpbb.com/customise/db/extension/subforumimages/support/topic/201456) and that we already had a ticket to address it here)

As we have access to the revision composer.json file, it's easy to add more details (like potentially any dependencies as @VSEphpbb mentioned in the ticket - which I'd like to do; does anyone know of any examples of extensions that depend on other extensions which I could use for testing?).

ch10

https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-626


VSEphpbb:
Here's a fake extension that you can use to test dependency. This just depends on viglink.
[foo.zip](https://github.com/phpbb/customisation-db/files/2852200/foo.zip)

styles/prosilver/template/common/contribution_details.html
```diff
@@ -63,6 +63,12 @@
 							<dd>{downloads.INSTALL_LEVEL}</dd>
 						</dl>
 						<!-- ENDIF -->
+						<!-- IF downloads.INSTALL_TO -->
```
VSEphpbb: Why are we using old PHPBB syntax instead of Twig? I think we should do Twig (in fact we should change it for the whole extension but that will be another PR).

battye: I thought about that too and also now with the old language function, but I also think it's best to tackle both of those things separately in their own PR's.

VSEphpbb: Yes separate PR (I'm working on that now). But lets please make all new code use Twig syntax 😄

battye:
Thanks for that download link, I've updated the code to include dependencies now - ignoring `php, composer/installers, phpbb/phpbb`.

Here's how it looks with that foo example:

ch11

styles/prosilver/template/common/contribution_details.html
battye: No problem, I've changed it to Twig

styles/prosilver/template/common/contribution_details.html
```diff
@@ -63,6 +63,18 @@
 							<dd>{downloads.INSTALL_LEVEL}</dd>
 						</dl>
 						<!-- ENDIF -->
+						{% if downloads.INSTALL_TO %}
+						<dl>
+							<dt>{{ lang('INSTALL_TO') }}{{ lang('COLON') }}</dt>
```
VSEphpbb: Instead of `}}{{` we should use `~` to concatenate two twig variables together

VSEphpbb: `{{ lang('INSTALL_TO') ~ lang('COLON') }}`

battye: Done

PR #287, opened 2019-05-15
Automatically generate the British English language pack

battye:
This is a console script which will allow the British English language pack to be automatically generated out of the zip package of phpBB.

**Usage:** the argument (with the `--name` flag) is the name of the phpBB zip that is added to the Titania code with each release. If the flag is omitted then you can specify the full path to a phpBB package that is not in Titania, if there was ever a reason to do that.

```
php bin/phpbbcli.php titania:extension:language phpBB-3.2.7.zip --name
```

@marc1706 said a little while ago on Discord that he doesn't have a script that does this but it might be useful. I think this could save time that would otherwise be spent manually generating the `en` package that we upload as a downloadable language pack.

@DavidIQ this should solve the problem discussed earlier about having to manually put the updated language pack in Titania. This script would just need to be run at the same time that [a new release is added](https://github.com/phpbb/customisation-db/commit/89b2a40b685f507dde8e2d337a4a1516ab82f0b7), with the auto-generated language pack zip being committed.

@Crizz0 Can you please check that [this example package I generated](https://github.com/battye/customisation-db/blob/923889942a0adbab0292fdc43d515b625ab88f15/includes/language_packages/british_english_3_2_7_auto_generated.zip) is a totally valid English language pack?

I'll leave this as a draft until Crizzo can confirm that the language pack format is correct, then I'll fix the naming in Git.


Crizz0:
Looks good, except the missing LICENSE file, we included in the last packages.

battye:
@Crizz0 I've made a change to include LICENSE.txt in the `language/en` directory. How does this look now?

Auto generated zip: [british_english_3_2_7.zip](https://github.com/phpbb/customisation-db/files/3193770/british_english_3_2_7.zip)

Crizz0:
The TV checks for LICENSE and not LICENSE.txt, so it will fail when we include a LICENSE.txt here with something like:
`Fatal in language/da/LICENSE.txt:  Missing required file`

You should include the LICENSE here. :)

battye:
Fixed at https://github.com/phpbb/customisation-db/pull/287/commits/53d002fd44d4c1dcf66b465357872a8d01dfc5a6

[british_english_3_2_7.zip](https://github.com/phpbb/customisation-db/files/3193882/british_english_3_2_7.zip)

Crizz0:
Look good.

PR #291, opened 2019-06-18
Skip EPV

battye:
Changes include:

* A checkbox for people with validation permission to bypass the EPV
  * If EPV is bypassed, this will be noted in a message in the queue post
* A message displayed informing users to contact a team member if they cannot submit their contribution

I want to test this a bit more before making the PR merge-able but also wanted to get it in the system early for review.

https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-735


language/en/contributions.php
```diff
@@ -153,7 +153,8 @@
 	'NEW_CONTRIBUTION'						=> 'New Contribution',
 	'NEW_REVISION'							=> 'New Revision',
 	'NEW_REVISION_READY'					=> 'New revision ready for processing.',
-	'NEW_REVISION_REPACK_COMPLETE'			=> 'New revision has been repacked to ensure packaging compliance. Click <em>Continue</em> to process through EPV.',
+	'NEW_REVISION_REPACK_COMPLETE'			=> 'New revision has been repacked to ensure packaging compliance. Click <em>Continue</em> to process through EPV. ',
+	'NEW_REVISION_REPACK_OVERSIZE'			=> 'If you are unable to submit your contribution, please contact a member of the <a href="https://www.phpbb.com/community/memberlist.php?mode=group&g=228811">Customisations Team</a>.',
```
DavidIQ: Let's not make this specific to our site. Maybe change to this:

> If you are unable to submit your contribution due to a system error, please contact us through your contribution's <a href="%s">validation discussion topic</a> for further assistance.

And add a link to the contribution's validation discussion topic. That's more open-ended for the user and does not make them decide who would be the one to speak to. Would need to verify that the validation discussion topic is actually created at contribution creation time as it might be getting created when creating a revision.

contribution/extension/type.php
```diff
@@ -139,6 +139,14 @@ public function perform_repack(\titania_contribution $contrib, \titania_revision
 		try
 		{
 			$this->repack($package, $contrib, $revision);
+			$repack_complete = $this->user->lang('NEW_REVISION_REPACK_COMPLETE');
+
+			// Oversized packages are over 2MB
+			if ($package->get_size() > 2097152)
+			{
+				$repack_complete .= $this->user->lang('NEW_REVISION_REPACK_OVERSIZE');
```
DavidIQ: There a way to only append this if the queue is enabled for the contribution type?

language/en/contributions.php
battye: Changed at https://github.com/phpbb/customisation-db/pull/291/commits/9cd178635cd6131b244878362717d05d75ec5990

Tested on a new contribution too, and it creates the topic before the revision.

contribution/extension/type.php
battye: How do you mean? Should it get to the point of repacking if the queue is not enabled?

DavidIQ: I might be mistaken, but I think this process runs independently from if the queue is enabled or not. If that's not the case then you can just ignore the comment :)

battye: How do you disable an individual queue?

battye:
I just realised that the Skip EPV option will appear on new revisions of any type, so I'll need to add some checking to make sure the type is an extension.

contribution/extension/type.php
DavidIQ: Nevermind this. I'm thinking the forum topic that gets all messages as well. This is fine as is.

battye:
Added the check against the contribution type, should be good to go now.

controller/contribution/revision.php
```diff
@@ -295,9 +298,14 @@ public function add($contrib_type, $contrib)
 			$this->cancel();
 		}
 
+		$this->skip_epv = $this->auth->acl_get('u_titania_mod_extension_validate') && (boolean) $this->request->variable('skip_epv', 0);
+		$can_skip_epv = $this->auth->acl_get('u_titania_mod_extension_validate') && $this->contrib->type instanceof \phpbb\titania\contribution\extension\type;
```
DavidIQ: The call to `request->variable` can just have the second parameter as `false` and then the casting isn't needed.

controller/contribution/revision.php
```diff
@@ -736,6 +744,9 @@ protected function process_steps()
 
 		$step = $steps[$step_num];
 
+		// Check if we are to skip the EPV
+		$this->skip_epv = $this->auth->acl_get('u_titania_mod_extension_validate') && (boolean) $this->request->variable('skip_epv', 0);
```
DavidIQ: The call to `request->variable` can just have the second parameter as `false` and then the casting isn't needed.

battye:
Type casts removed

PR #294, opened 2019-09-15
ATOM Feeds

battye:
This change allows for:

* An overall feed of contribution revisions which will track any new releases (`/db/feed`)
* A specific feed which will track only new releases of a single contribution (`/db/extension/foo_bar_ext/feed`)

The restrictions are that the feeds will not appear if for the board itself feeds are disabled, and that contributions which only appear in hidden categories will not appear in the feeds. A screenshot of how the overall feed is below:

ch20

https://tracker.phpbb.com/browse/CUSTDB-786


includes/overlords/contribs.php
```diff
@@ -570,4 +570,167 @@ public static function build_sort()
...
+				$rows[] = $feed_rows;
+			}
+		}
+
+		phpbb::$db->sql_freeresult($result);
+
+		if (!$rows)
+		{
+			// If there's no results, we can't proceed
+			trigger_error('FEED_CDB_NOT_AVAILABLE');
```
kasimi: If I understand it correctly, for example, when a contribution's only validated revision is pulled (e.g. for security reasons), is the feed supposed to throw an error? What's the expected behavior? Should we render a valid but empty feed here instead?

styles/prosilver/template/common/contribution_details.html
```diff
@@ -96,8 +95,11 @@
 								{% if IS_SUBSCRIBED %}{{ lang('UNSUBSCRIBE') }}{% else %}{{ lang('SUBSCRIBE') }}{% endif %}
```
kasimi: Just an idea, bit more compact: `{{ lang(IS_SUBSCRIBED ? 'UNSUBSCRIBE' : 'SUBSCRIBE') }}`

battye: Done

includes/overlords/contribs.php
battye: That's correct, the feed is only served for contributions that have valid revisions.
But it's easy to change it to an empty feed if that behaviour would be preferred? @DavidIQ @VSEphpbb @kasimi

DavidIQ: I think that would be preferred, yes. No sense in triggering and likely logging an error if it's just someone following something they shouldn't.

battye: Done - I've changed it at e61c13fe90baf6821bbf65557e78f2a50d62cc09 so an empty feed will be displayed now.

includes/overlords/contribs.php
```diff
@@ -570,4 +570,162 @@ public static function build_sort()
...
+				// Make the name including the version
+				$feed_rows['item_title'] = $row['contrib_name'] . ' ' . $row['revision_version'];
+
+				if ($row['revision_name'])
+				{
+					// Include the code name if it's supplied
+					$feed_rows['item_title'] .= ' (' . $row['revision_name'] . ')';
+				}
+
+				$feed_rows['item_author'] = $row['username_clean'];
```
3D-I: I left a comment here: https://github.com/phpbb/customisation-db/commit/ea1d151fc755175e78914de30fc8a3160ada846e#r35146186

`$row['username']` should be used instead.

PR #296, opened 2019-09-19
Fix ATOM feed author username

3D-I:
As of now `3Di` renders as `3di` in the feed.
```

    <entry>
        <author><name><![CDATA[3di]]></name></author>
```


battye:
The reason I used `$row['username_clean']` is because I wasn't sure how feed viewers would handle weird characters. But I don't use feeds so it might be a non-issue, and if so then I have no objection to this changing to `$row['username']` :)

DavidIQ:
I think it's fine since the content header is set to atom/rss. If the client can't handle special characters I'd think there's more issues it would have beyond the username. 😀

PR #297, opened 2019-09-25
[CUSTDB-727] Featured contributions on the main index in Titania

battye:
This change addresses [CUSTDB-727](https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-727), which highlighted the benefits of having featured contributions appear on the index of Titania.

* Two featured contributions will appear on the index; each day different items will be randomly selected
  * One will be a historical contribution
  * One will be a recent contribution (from one of the last five new and approved contributions)
* When the user moves away from the main index, the featured contributions container will slide away and be hidden

captured

https://tracker.phpbb.com/projects/CUSTDB/issues/CUSTDB-727


styles/prosilver/template/index_body.html
```diff
@@ -2,7 +2,46 @@
...
+					{% if contribs.PREVIEW %}
+					<img src="{{ contribs.PREVIEW }}" alt="{{ contribs.CONTRIB_NAME }}" class="screenshot" />
+					{% else %}
+					<img src="{{ T_TITANIA_THEME_PATH }}/images/types/{{ contribs.CONTRIB_TYPE_CLEAN }}.png" alt="{{ lang('NO_SCREENSHOT') }}" class="screenshot" />
+					{% endif %}
+				</div>
+			</div>
+			{{ contribs.CONTRIB_RATING_STRING }}
+			<br />
+			{{ lang('BY') }} {{ contribs.AUTHOR_NAME_FULL }}<br /><br />
```
DavidIQ: Would rather not use `br` tags. Can you try using `p` tags around the above instead and around the for loop that follows?

includes/overlords/contribs.php
```diff
@@ -301,6 +313,44 @@ public static function display_contribs($mode, $hierarchy_ids, $branch = false,
...
+
+					'WHERE'	=> 'r.revision_status = ' . ext::TITANIA_REVISION_APPROVED . '
+						AND r.revision_submitted = 1
+						AND c.contrib_status = ' . ext::TITANIA_CONTRIB_APPROVED . '
+						AND r.contrib_id = c.contrib_id',
+
+					'ORDER_BY'	=> 'c.contrib_id ASC',
+				];
+
+				$sql = phpbb::$db->sql_build_query('SELECT_DISTINCT', $sql_ary);
```
DavidIQ: Will this include MODs which we aren't displaying anymore?

battye: That's a good point. I might also see if we can add some extra filtering on it so it will only pull from the latest phpBB version, so we don't end up with something like a 3.0 style making the featured list.

battye:
Done

controller/index.php
```diff
@@ -399,11 +399,18 @@ protected function list_contributions($categories, $sort_url)
...
+
 		$data = \contribs_overlord::display_contribs($mode, $categories, $branch, $sort, 'contribs', $this->status);
 
 		// Canonical URL
 		$data['sort']->set_url($sort_url);
-		$this->template->assign_var('U_CANONICAL', $data['sort']->build_canonical());
+
+		$this->template->assign_vars([
+			'U_CANONICAL' => $data['sort']->build_canonical(),
+			'U_IS_TITANIA_INDEX' => (!$sort_url) ? true : false,
```
VSEphpbb: Please just make this: `'S_IS_TITANIA_INDEX' => empty($sort_url),` or `'S_IS_TITANIA_INDEX' => $sort_url === 0,`

It's a boolaean switch so we should use the `S_` prefix. Also, it's better to just have a boolean condition rather than the whole if it strue or false then make it true or false conditional statement.

language/en/common.php
```diff
@@ -128,6 +128,7 @@
 	'EDIT_CATEGORY'				=> 'Edit Category',
 	'ERROR'						=> 'Error',
 
+	'FEATURED_CONTRIBUTION'		=> 'Featured contribution',
```
VSEphpbb: I'm just not sure about using the word "Featured". Since these are being selected totaly at random, they're not really _featured_, if that makes any sense. When I think of featured I think of us intentionally picking a contrib that we want to prominently feature, based on its merits. I'd almost prefer calling this something like:

"Random Contribution Of The Day" or maybe just "Contribution Of The Day" or if anybody has a better thought. 

Or maybe: "Recent Contribution of the Day" and "Historic Contribution of the Day"


Again my words suck but the germ of an idea is there I think.

battye: It might just be overcomplicating it if we change the wording. I was just thinking "featured" in the sense that it's featured prominently on the main page.

controller/index.php
battye: Done

battye:
@DavidIQ Are there any other changes required for this PR?

DavidIQ:
Far as I'm concerned this looks good to go. Anything else @VSEphpbb ?

VSEphpbb:
@DavidIQ the code is fine. I just still have a conceptual issue with this. In the Ticket, @michaelcullum says he wanted this to feature popular and/or new contribs. However I think this seems to be totally random. I'd rather see this have a bias towards new submits and popular/lotta downloads contribs.

DavidIQ:
I think I like the idea presented here though, but maybe it should be changed to say it's a random contribution and not featured. In connection with the ticket perhaps there can be "Most Downloaded", "Highest Rated", "Newest" sections, but that should be done as a new pr

battye:
@DavidIQ Done - changed the wording to random.

I can do a second PR for this that will put bias on contributions with high downloads, ratings (once fixed) etc.

DavidIQ:
Perfect, thanks

PR #301, opened 2019-12-20
Move custom config file settings into Administration

VSEphpbb:
Well this took over a year, lol.

This would move Titania's root config.php configuration options into its Administration panel. 

If this were added to phpbb.com, all you would need to do is go to the Configuration settings of Titania’s admin panel, and hit submit. That would store all the existing settings (that are currently in a custom config.php file) into the database, and then it would be safe to delete the custom config.php file and the example.config.php file.

Here's an image:
sample3

This should finalize the task of eliminating the need for separate phpbb-com branches here that have modified versions of titania for use on phpbb.com.

Also note: Not ALL of the config options are currently shown in this panel, because some just shouldn't really ever need changing. But they can be shown here just by uncommenting them and creating lang key/vars for them. The config options currently shown here are all the one's that phpbb.com has needed to configure (per its custom config.php file).

Lastly note that titania/config/config.php is still the master configuration file, upon which this admin panel depends (this panel is only to replace the custom config.php file, based off the example.config.php). So the master config.php is still what would be updated to add new phpBB branches.


VSEphpbb:
Made some twekas to clean up how it looks a little better, and added a few more reasonable config settings for the few other people in the world trying to use this.

DavidIQ:
I'll try this branch out on my local copy of .com this week hopefully.

VSEphpbb:
@DavidIQ cool. Like I said above, when you load this page, all the current settings should already be there. You just need to hit Submit to actually get them saved into the config table. Then you can nuke the custom config.php file and everything should still work.

styles/prosilver/template/manage/config_settings.html
```diff
@@ -0,0 +1,63 @@
...
+							{% elseif options.TYPE === 'bool' %}
+								<input type="radio" name="{{ options.NAME }}" value="1" {% if options.VALUE %}checked="checked"{% endif %} /> {{ lang('YES') }}
+								<input type="radio" name="{{ options.NAME }}" value="0" {% if not options.VALUE %}checked="checked"{% endif %} /> {{ lang('NO') }}
+							{% elseif options.TYPE === 'array|string' %}
+								{% for subKey, subValue in options.VALUE %}
+									<label for="{{ options.NAME }}_{{ subKey }}"><input type="text" name="{{ options.NAME }}_{{ subKey }}" id="{{ options.NAME }}_{{ subKey }}" size="50" value="{{ subValue }}" class="inputbox autowidth"> {{ lang('TI_KEY_' ~ subKey) }}</label><br />
+								{% endfor %}
+							{% elseif options.TYPE === 'array|int' %}
+								{% for subKey, subValue in options.VALUE %}
+									<label for="{{ options.NAME }}_{{ subKey }}"><input type="number" name="{{ options.NAME }}_{{ subKey }}" id="{{ options.NAME }}_{{ subKey }}" min="0" max="9999999999" value="{{ subValue }}" class="inputbox autowidth"> {{ lang('TI_KEY_' ~ subKey) }} </label><br />
```
DavidIQ: Translations are set to `-1` so having `min="0"` here causes the validation to fail and can't save the form. Needs to be adjusted to `-1` or removed.

VSEphpbb: Updated

PR #302, opened 2019-12-28
Prepare for 3.3 release

marc1706:


VSEphpbb:
Should there be a 3.3 branch created?

marc1706:
Good question. Yes but no. So this only adds support for submitting 3.3 extensions but it does not add support for phpBB 3.3 to Titania. .com will first move to having 3.3 submissions before actually moving over to 3.3. ;)

VSEphpbb:
@marc1706 I actually meant a Git Branch

DavidIQ:
Not needed right now. Would have to also change the website update script for that

marc1706:
I'd move to a 3.3.x branch when it's actually meant to run on phpBB 3.3.x

PR #306, opened 2020-01-01
WIP: Think this needs updating for 3.3 too

VSEphpbb:
But not totally sure, lol.


paul999:
But only once we run phpBB. Com on 3.3, which we don't plan initially

battye:
The symfony-2.8 branch is the PHP5 compatible branch, so we can't change that just yet.

DavidIQ:
This should go against master, not 3.2.

PR #308, opened 2020-01-03
Fixing some code errors and cleaning out old stuff

VSEphpbb:
Some incredibly outdated docs have been removed and/or updated.

Also, I’m shocked Titania actually works at all. There are so many inspection errors, and so much spagetti code I fixed only the simplest and most obvious errors that shouldn’t appear to break anything (all the other things I’d like to fix will probably break something unforeseen).


controller/ucp/subscriptions.php
```diff
@@ -184,7 +184,7 @@ protected function display_sections()
 					if (!$row['contrib_id'])
 					{
 						$this->delete_subscription($row['watch_object_type'], $row['watch_object_id'], false);
-						continue;
```
paul999: Please make sure this works on 7.3

docs/SETUP.md
```diff
@@ -2,13 +2,27 @@
 
 This is a quick setup guide to walk you through the process of installing and configuring Titania.
 
+## Requirements
+
+1. phpBB 3.1.x, 3.2.x or 3.3.x
```
paul999: you are sure?

controller/ucp/subscriptions.php
VSEphpbb: Yes. In fact 7.3 would issue a warning here saying we should be using `continue 2`

docs/SETUP.md
VSEphpbb: Just going by Titania's composer.

https://github.com/phpbb/customisation-db/blob/3.2.x/composer.json#L74

controller/ucp/subscriptions.php
paul999: Ok, just wanted to be sure :)

PR #310, opened 2020-01-08
Get latest lang pack of the selected branch

battye:
This should fix the issue @Crizz0 mentioned of it always picking the 3.3.0 version (because it was the most recent), even if someone selected the 3.2 branch.

I did a test of this and I got different validation results based on the version I selected, so it appears to be performing as expected. eg:

_Fatal: 69, Error: 7, Warning: 39, Notice: 53_ for 3.3
_Fatal: 19, Error: 7, Warning: 35, Notice: 53_ for 3.2

### DIFF
--- a/contribution/translation/prevalidator.php
+++ b/contribution/translation/prevalidator.php
@@ -72,7 +72,7 @@ class prevalidator
 	 * Get the latest British English pack
 	 * @return array Return the path and version to the pre-supplied British English language pack
 	 */
-	private function get_latest_english_pack()
+	private function get_latest_english_pack($selected_branch = null)
 	{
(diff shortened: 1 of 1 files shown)


VSEphpbb:
I can't reproduce whatever @Crizz0 was seeing

- I add a new revision on the 3.2 branch and it goes to 3.2.9...
- I add a new revision on the 3.3 branch and I get the error "Invalid phpbb-version, allowed versions: 3.2"
- I add a new revision to both the 3.2 and 3.3 branch and it goes to both 3.2.9 and 3.3.0

My results are the same with or without this patch

VSEphpbb:
Seems to work.

PR #313, opened 2020-02-21
Use secure connection for extension version check

kasimi:


contribution/extension/type.php
```diff
@@ -336,6 +336,7 @@ protected function set_version_check(array $data, \titania_contribution $contrib
 				'host'		=> $parts['host'],
 				'directory' => $directory,
 				'filename'	=> substr($parts['path'], strlen($directory) + 1),
+				'ssl'		=> true,
```
DavidIQ: We'll probably want to go off of what's in `$parts`. Looks like it would be in `$parts['scheme']` so maybe something like:
```
'ssl' => isset($parts['scheme']) ? $parts['scheme'] === 'https' : true,
```
We'll just default to `true` if it isn't defined.

kasimi: Isn't this unnecessarily complex? .com going back to `http://` is unlikely, and if it does, existing extension installations will still have `'ssl': true` in their version check because `https://` was used at the time the revision was approved.

DavidIQ: This isn't designed just for use on .com though and is the reason for the suggestion.

kasimi: Ah that makes sense 👍

PR #327, opened 2021-03-15
Fix `.contrib-quickview`

Dark1z:
The CSS class `.contrib-quickview` has some visual bugs which leaves some empty blocks, which is Fixed.

Best regards 👍


Dark1z:
For Example in `Customisation Database` > `Author` > `Contributions`:
Before Fix: 

After Fix:

Dark1z:
For Example in `Customisation Database` > `Extensions`:
Before Fix: 

After Fix:

DavidIQ:
Thanks for the fix 👍

PR #330, opened 2021-05-03
Store array based configs in config_text table

iMattPro:
So with this, you would need to go to the Administration->Configuration Settings and just submit the form once. That will restore settings, and move the more complex array based settings from config to config_text.


marc1706 (approved):

marc1706:
We can test that on the dev server before deploying it to .com

DavidIQ (approved):

PR #342, opened 2022-03-27
[CUSTDB-824] Add gender to the contrib status according to the contrib type

Skouat:
https://tracker.phpbb.com/browse/CUSTDB-824


DavidIQ:
I'm confused about the title of this MR. How is this doing anything with gender? Maybe something is being lost in translation here...

Skouat:
Hi @DavidIQ

Perhaps, the title is not the best.

IMO, nothing is lost in translation.
But I understand your confusion. For English speakers, it could make the translation more "complex".
But for the Latin language, you have to agree the verb with the subject.
For exemple, in Latin languages, "style" is a masculine noun and "extension" is a feminine noun.

DavidIQ:
Yes I am perfectly aware of the difference and the rules for latin languages (my primary language is Spanish). Problem is that since English is the primary language the changes didn't seem to show any difference. And since no screenshot was provided of what these changes actually do it's even less clear.

Skouat:
> Problem is that since English is the primary language the changes didn't seem to show any difference.

Indeed, this change doesn't show a difference for English language.
This is only for other languages.
And even if English is the primary language, this doesn't prevent us from translating Titania into another language...

> And since no screenshot was provided of what these changes actually do it's even less clear.

Have you take a look on Jira? The link is in the first message of this PR.

DavidIQ:
Got it, thanks. 

/merge

DavidIQ (approved):

PR #350, opened 2022-07-31
Fixing stuff that won’t work with PHP 8

iMattPro:


Skouat:
Hi @iMattPro 
What do you think about removing the 'safe_mode' code in [attachment/attachment.php](https://github.com/phpbb/customisation-db/blob/3.3.x/attachment/attachment.php#L356) (feature removed on PHP >=5.4)
And update PHP version to 7.1.3 in composer.json?

Skouat:
Hi @iMattPro ,
There is another issue with PHP 7.4 and newer.
>/ext/phpbb/titania/includes/overlords/attention.php on line 233: Trying to access array offset on value of type bool

This happens when there is a report on attention tab.

PR #353, opened 2022-08-30
fix: missing "id" param for phpbb.titania.queue.item route when using Sphinx

Skouat:
This fix ticket https://github.com/phpbb/customisation-db/issues/369
And some other stuff


DavidIQ:
Can you give this a better title please? Thanks.

PR #385, opened 2024-01-29
Fix link for tag in breadcrumbs

battye:
@DavidIQ Can I confirm I understand correctly what you had raised on chat?

This fix ensures that in the validation queue, if you go into a particular type (e.g., extensions) and then click on one of the subcategories (e.g., a tag like "New"), when you click on the queue item (e.g., Digital Clock in the example below) the breadcrumbs were generating with the wrong link.

Previously, the tag ("New") was linking to the item instead of back the queue type.

With this change, the tag ("New") in the breadcrumb links back to the extensions type list with the tag to filter it by New.

Screenshot 2024-01-29 at 4 34 34 pm

### DIFF
--- a/controller/manage/queue/item.php
+++ b/controller/manage/queue/item.php
@@ -85,6 +85,11 @@ class item extends \phpbb\titania\controller\manage\base
 			return $this->helper->needs_auth();
 		}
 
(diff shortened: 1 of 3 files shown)


DavidIQ:
Looks good to me, exactly what I had reported. Thanks!

Issue #365, opened 2023-09-23
Remove MOD and MPV code


battye:
MODs and MPV are no longer accepted into the CDB. Code related to it should be removed.

Reporter: DavidIQ
Created: 09/Sep/17 10:45 AM

PR #400, opened 2025-05-31
Remove automod/MPV

Skouat:
This PR handles the issue #365


iMattPro (approved):

iMattPro:
This makes me think we should also remove "Zend" as a search engine, then that would allow us to delete the `zend` and `ezcomponents` junk in includes/library

Skouat:
> This makes me think we should also remove "Zend" as a search engine, then that would allow us to delete the `zend` and `ezcomponents` junk in includes/library

If you confirm that both can be removed, I will propose a pull request for that.

iMattPro:
I can neither confirm nor deny. But I know on dot-com we are not using the "zend" search engine option. Not sure what maybe @DavidIQ thinks.

DavidIQ:
> I can neither confirm nor deny. But I know on dot-com we are not using the "zend" search engine option. Not sure what maybe @DavidIQ thinks.

Should be fine 👍

Skouat:
> Should be fine 👍

Good. I will check this.

And how about this PR? Is it good for merging?

Skouat:
Hi @DavidIQ 
Same than the other PR…
Are there any other changes to apply? Or is it okay for merging?
Thanks

PR #406, opened 2025-09-30
Composer repository builder should also make a phpBB4 only packages.json

iMattPro:
@DavidIQ and @marc1706 This PR would be in addition to https://github.com/phpbb/phpbb/pull/6873

This PR updates the composer repo build for phpBB4’s extension catalog so that only extensions marked as supporting the phpBB 4 branch get included. Basically, it cuts down the risk of presenting extensions that aren’t submitted & validated for phpBB4 and could break boards.

### DIFF
--- a/composer/repository.php
+++ b/composer/repository.php
@@ -15,6 +15,7 @@ namespace phpbb\titania\composer;
 
 use phpbb\config\config;
 use phpbb\exception\runtime_exception;
+use phpbb\titania\ext;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder;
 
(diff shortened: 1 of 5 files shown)


DavidIQ:
I thought these changes weren't going to be needed. Not sure this will be ok. Like I said there is at least one extension that makes use of this data so not sure we want to break their stuff.

iMattPro:
Until some user submits an extension that is for phpBB3, and not compatible with phpBB4, yet they updated their composer installers to 2.0 for some reason. Then their extension could appear in phpBB4 forums as installable.

DavidIQ:
This is the extension I'm referring to that is referencing the composer repo in Titania: https://github.com/BoardTools/upload/blob/develop-3.2.0/includes/sources/extensions_list.php#L18

These changes would essentially break this extension. Whether or not it is still active is another story.

iMattPro:
> This is the extension I'm referring to that is referencing the composer repo in Titania: https://github.com/BoardTools/upload/blob/develop-3.2.0/includes/sources/extensions_list.php#L18
> 
> These changes would essentially break this extension. Whether or not it is still active is another story.

Ah I forgot about that extension.

OK, here's an alternative approach then. Let's generate a second list that is for phpBB4 branch submissions only, in addition to the standard all extensions list.

This way, the extension(s) currently using our repository list can continue unaffected, and in phpBB4's extension catalog we instead have it download its repository packages from the phpbb4 only branch, aka: `/db/composer/packages-40.json`

iMattPro:
I added a new URL path so phpBB would need to update it's composer repository config setting to point to:
`https://www.phpbb.com/customise/db/composer/40/` to only load extensions submitted to CDB that are supprt the 4.0.x branch, ie:

Screenshot 2025-09-30 at 4 47 57 PM

What gets generated when the repo-rebuilder runs, looks somewhat like this (regular packages.json files for all extensions, and `40/` sub-dir with json files just for phpBB4 extensions:

Screenshot 2025-10-02 at 12 46 27 PM

DavidIQ (approved):

DavidIQ:
FYI @battye since the next version of the CDB needs to also support this.

PR #446, opened 2026-07-31
Limit the demo box to branches with an approved revision

### LINKED ISSUE #377: Styles demo install box showing for non-supported versions
A style that has only been approved for 3.3 is showing the install box for 3.2:

image

image

Should be going by what version(s) the style has been approved for.

ECYaz:
Fixes #377

The manage page built the demo URL box from every branch open for uploads, so a style approved only for 3.3 still offered the 3.2 field and install button. The demo box and the details page demo menu are now built from the branches that actually have an approved revision, the same source the download box uses. Stored URLs for branches that are not shown survive a save, so a pulled revision hides its demo without losing the URL if it is approved again.

### DIFF
--- a/includes/objects/contribution.php
+++ b/includes/objects/contribution.php
@@ -1036,6 +1036,38 @@ class titania_contribution extends \phpbb\titania\entity\message_base
 	}
 
 	/**
+	* Get the branches for which the contribution has an approved revision.
+	*
+	* @return array Branch names limited to the branches that have an approved
+	*	revision, keyed by branch - example: 31 => 'phpBB 3.1.x'
+	*/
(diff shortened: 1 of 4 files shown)


includes/objects/contribution.php
```diff
@@ -1041,16 +1041,58 @@ public function get_url($page = '', $parameters = array())
 	* @return array Branch names limited to the branches that have an approved
 	*	revision, keyed by branch - example: 31 => 'phpBB 3.1.x'
 	*/
-	public function get_approved_branches()
+public function get_approved_branches()
```
iMattPro: Indentation is off

